PR #7, opened 2023-01-17
PSR-43 created test cases for views and forms

narendrapaduchuri:


app/forms/DatePageFormProvider.scala
```diff
@@ -0,0 +1,41 @@
...
+
+package forms
+
+import forms.mappings.Mappings
+import play.api.data.Form
+
+import java.time.LocalDate
+import javax.inject.Inject
+
+class DatePageFormProvider @Inject() extends Mappings {
```
kurtlogan, line 25: I was thinking that maybe the provider could return a case class that wraps the form and the view can accept that as a parameter. This would mean that a dev couldn't create a form with the wrong fields but still satisfys the type signature. This will need some thought as form as methods that are needed and would break it were to be a parameter on another case class. Perhaps we can go through this

app/viewmodels/models/DatePageViewModel.scala
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package viewmodels
+
+import play.api.i18n.Messages
+import play.api.mvc.Call
+
+case class DisplayMessage(key: String, args: List[Any]) {
```
kurtlogan, line 22: Lets pull display message into it's own class

test/controllers/SampleTestSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+class SampleTestSpec extends AnyWordSpec with Matchers{
```
kurtlogan, line 22: Is this file needed?

test/views/PensionSchemeViewSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+    implicit val request = FakeRequest()
+    implicit val mess = messages(app)
+
+    "test 1" in {
+
+      val form = new PensionSchemeForm()
+      val viewModel = PensionSchemeViewModel(
+        DisplayMessage("value 1"),
+        DisplayMessage("header"),
+        //DisplayMessage("value 3")
```
kurtlogan, line 48: Don't comment out code, remove it instead

test/views/PensionSchemeViewSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+
+
+  running(_ => applicationBuilder()) { implicit app =>
+
+    val view = app.injector.instanceOf[PensionSchemeView]
+
+    implicit val request = FakeRequest()
+    implicit val mess = messages(app)
+
+    "test 1" in {
```
kurtlogan, line 42: Need a better test name

test/views/PensionSchemeViewSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+        DisplayMessage("value 1"),
+        DisplayMessage("header"),
+        //DisplayMessage("value 3")
+      Call("GET", "/value")
+      )
+
+   val failingForm = form("value")
+
+      val result = view(failingForm, viewModel)
+      Jsoup.parse(result.toString()).body(). toString must include("header")
```
kurtlogan, line 55: no space between . and toString

test/views/PensionSchemeViewSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+import play.api.test.FakeRequest
+import play.api.test.Helpers.running
+import viewmodels.models.PensionSchemeViewModel
+import viewmodels.DisplayMessage
+import views.html.PensionSchemeView
+import org.jsoup.Jsoup
+import play.api.mvc.Call
+
+
+class PensionSchemeViewSpec extends SpecBase {
```
kurtlogan, line 32: More tests needed in this file to cover success and failure cases

westwater: Added a success and failure case 👍

app/forms/DatePageFormProvider.scala
westwater: Yeah it would be good to go through this

narendrapaduchuri: Yes, we can go through this

test/views/PensionSchemeViewSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    implicit val request = FakeRequest()
+    implicit val mess = messages(application)
+
+    "PensionsSchemeView" - {
+
+      "view should render correctly" in {
+
+        val form = new PensionSchemeForm()
+        val viewModel = PensionSchemeViewModel(
```
kurtlogan, line 47: You need to test that every value passed into the view model is rendered on the page. So title should be tested, heading should be tested and the onSubmit should also be tested

test/views/PensionSchemeViewSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+          title = DisplayMessage("testTitle"),
+          heading = DisplayMessage("testHeading"),
+          Call("GET", "/value")
+        )
+
+        val failingForm = form("value").bind(Map("failing" -> "data"))
+
+        val result = view(failingForm, viewModel)
+        Jsoup.parse(result.toString()).body().getElementById("error-summary-title").text() mustBe "There is a problem"
```
kurtlogan, line 69: 2 error messages appear on the page, both should be tested

app/viewmodels/models/PensionSchemeViewModel.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package viewmodels.models
+
+import play.api.mvc.Call
+import viewmodels.DisplayMessage
+
+case class PensionSchemeViewModel(
```
kurtlogan, line 22: Might be a little clunky creating a view modeland having to pass `DisplayMessage`. Create a new apply method in the companion object with a signature `def apply(title: String, heading: String, onSubmit: Call): PensionSchemeViewModel` that applies the string values to `DisplayMessage`

app/forms/DatePageFormProvider.scala
westwater: Had a chat and agreed to leave this for now

test/views/PensionSchemeViewSpec.scala
westwater: Does the title get passed to the html? I thought it was for the browser tab

westwater (approved):

PR #10, opened 2023-01-24
Psr 71

kurtlogan:


narendrapaduchuri (dismissed):

app/repositories/SessionRepository.scala
```diff
@@ -54,14 +55,14 @@ class SessionRepository @Inject()(
...
-  def keepAlive(id: String): Future[Boolean] =
+  def keepAlive(id: String): Future[Unit] =
     collection
       .updateOne(
         filter = byId(id),
         update = Updates.set("lastUpdated", Instant.now(clock)),
       )
       .toFuture
-      .map(_ => true)
+      .as(())
```
westwater, line 65: minor: do you think it's worth adding an extra function to FutureOps to handle this case? (like .unit in cats)

app/viewmodels/models/ContentPageViewModel.scala
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package viewmodels.models
+
+import play.api.mvc.Call
+import viewmodels.DisplayMessage
+
+case class ContentPageViewModel(
```
westwater, line 22: nice abstraction 👍

westwater (approved):

narendrapaduchuri (approved):

PR #12, opened 2023-01-30
PSR-31 Added check your scheme details page

westwater:


conf/application.conf
```diff
@@ -72,6 +72,12 @@ microservice {
           host = localhost
           port = 8203
       }
+
+      managePensions {
```
kurtlogan, line 76: Move this to urls section, services are internal so host will change to manage-pensions.public.mdtp which isn't accessible from outside mdtp, put this in the urls section instead

kurtlogan (approved):

kurtlogan (dismissed):

Vamshi-Muniganti (approved):
looks good to me

PR #14, opened 2023-02-03
Appended srn to internal id when persisting to session cache DB

westwater:


app/controllers/actions/DataCreationAction.scala
```diff
@@ -30,7 +30,8 @@ class DataCreationActionImpl @Inject()(sessionRepository: SessionRepository)(imp
 
     request.userAnswers match {
       case None =>
-        val userAnswers = UserAnswers(request.getUserId)
+        val userAnswersKey = request.getUserId + request.request.schemeDetails.srn
```
kurtlogan, line 33: Could you create a method on `OptionalDataRequest` so you can get he scheme details from it without having to call `request.request`

kurtlogan (approved):

kurtlogan (approved):

PR #18, opened 2023-02-15
PSR-87 Added bank account summary page

westwater:


app/config/Constants.scala
```diff
@@ -26,4 +26,6 @@ object Constants {
 
   val delimitedPSA = "DELIMITED_PSAID"
   val detailsNotFound = "no match found"
+
+  val MaxSchemeBankAccounts = 10
```
kurtlogan, line 30: Values start with lowercase letters

app/config/Binders.scala
```diff
@@ -29,4 +31,17 @@ object Binders {
 
     override def unbind(key: String, value: Srn): String = value.value
   }
+
+  implicit val max10PathBinder: PathBindable[Max10] =
```
kurtlogan, line 35: This can be made more generic to handle any Refined[Int, ?], just need an instance of Validate[Int, ?] in implicit scope
```
  implicit def refinedIntPathBinder[T](implicit ev: Validate[Int, T]): PathBindable[Refined[Int, T]] =
    new PathBindable[Refined[Int, T]] {
      override def bind(key: String, value: String): Either[String, Refined[Int, T]] = value
        .toIntOption
        .toRight(s"value for key $key was not an Integer")
        .flatMap(refineV[T](_))

      override def unbind(key: String, value: Refined[Int, T]): String = value.value.toString
    }

  implicit def refinedIntJSLiteral[T]: JavascriptLiteral[Refined[Int, T]] =
    (value: Refined[Int, T]) => value.value.toString
```

app/models/UserAnswers.scala
```diff
@@ -63,6 +65,9 @@ final case class UserAnswers(
         page.cleanup(None, updatedAnswers)
     }
   }
+
+  def schemeBankAccounts(srn: Srn): List[BankAccount] =
```
kurtlogan, line 69: This class should not really contain any logic to any specific page

One thing you could do is have an object Page as well as a case class Page that onlt implements the `Getter[A]` and returns the whole collection rather than a single element. Somthing like
```
case class SchemeBankDetailsPage(srn: Srn, index: Int) extends QuestionPage[BankAccount] {
    def path: JsPath = JsPath \ toString \ index
    def toString: String = SchemeBankDetailsPage.pageName
}

object SchemeBankDetailsPage extends Gettable[List[BankAccount]] {
    val pageName: String = ???
    def path: JsPath = JsPath \ pageName 
}
```

app/navigation/Navigator.scala
```diff
@@ -16,29 +16,43 @@
...
       if(ua.get(page).contains(true))
-        routes.SchemeBankAccountController.onPageLoad(srn, NormalMode)
+        routes.SchemeBankAccountController.onPageLoad(srn, refineMV[OneToTen](1), NormalMode)
       else
         routes.UnauthorisedController.onPageLoad
-
-    case SchemeBankAccountPage(srn) => _ => routes.UnauthorisedController.onPageLoad
-    case _                          => _ => routes.IndexController.onPageLoad
+    case SchemeBankAccountPage(srn) => _ => routes.SchemeBankAccountSummaryController.onPageLoad(srn)
+    case SchemeBankAccountSummaryPage(srn, addBankAccount @ true) => userAnswers =>
```
kurtlogan, line 44: `addBankAccount @` isn't needed

app/navigation/Navigator.scala
```diff
@@ -16,29 +16,43 @@
...
+      val numBankAccounts = userAnswers.get(SchemeBankAccountPage(srn)).toList.flatten.length
+      refineV[OneToTen](numBankAccounts) match {
+        case Left(_)                               => routes.SchemeBankAccountSummaryController.onPageLoad(srn)
+        case Right(Refined(MaxSchemeBankAccounts)) => routes.UnauthorisedController.onPageLoad
+        case Right(index)                          => refineV[OneToTen](index.value + 1) match {
+          case Left(_)          => routes.UnauthorisedController.onPageLoad
+          case Right(nextIndex) => routes.SchemeBankAccountController.onPageLoad(srn, nextIndex, NormalMode)
+        }
+      }
+    case SchemeBankAccountSummaryPage(_, addBankAccount @ false) => _ => routes.UnauthorisedController.onPageLoad
```
kurtlogan, line 54: Same as above

app/navigation/Navigator.scala
```diff
@@ -16,29 +16,43 @@
...
-
-
   private val normalRoutes: Page => UserAnswers => Call = {
     case StartPage(srn)             => _ => routes.SchemeDetailsController.onPageLoad(srn)
     case SchemeDetailsPage(srn)     => _ => routes.CheckReturnDatesController.onPageLoad(srn, NormalMode)
 
     case page@CheckReturnDatesPage(srn)  => ua =>
       if(ua.get(page).contains(true))
-        routes.SchemeBankAccountController.onPageLoad(srn, NormalMode)
+        routes.SchemeBankAccountController.onPageLoad(srn, refineMV[OneToTen](1), NormalMode)
```
kurtlogan, line 40: Should this alaways go to the first bank account? Maybe if there is data it should go to the list page. Maybe ask Josh or Michael

app/navigation/Navigator.scala
```diff
@@ -16,29 +16,43 @@
...
+    case SchemeBankAccountSummaryPage(srn, addBankAccount @ true) => userAnswers =>
+      val numBankAccounts = userAnswers.get(SchemeBankAccountPage(srn)).toList.flatten.length
+      refineV[OneToTen](numBankAccounts) match {
+        case Left(_)                               => routes.SchemeBankAccountSummaryController.onPageLoad(srn)
+        case Right(Refined(MaxSchemeBankAccounts)) => routes.UnauthorisedController.onPageLoad
+        case Right(index)                          => refineV[OneToTen](index.value + 1) match {
+          case Left(_)          => routes.UnauthorisedController.onPageLoad
+          case Right(nextIndex) => routes.SchemeBankAccountController.onPageLoad(srn, nextIndex, NormalMode)
+        }
+      }
```
kurtlogan, line 53: I think this can be shortend
I don't think there is a difference between having the max number already and exceeding the max after we increment it so it could become
```suggestion
      refineV[OneToTen](numBankAccounts + 1) match {
        case Left(_)                  => routes.SchemeBankAccountSummaryController.onPageLoad(srn)
        case Right(nextIndex) => routes.SchemeBankAccountController.onPageLoad(srn, nextIndex, NormalMode)
      }
```

kurtlogan (changes requested):

app/config/Constants.scala
westwater: It's uppercase so I can use it in a pattern match statement without prefixing with `Constants.` but can change it to lowercase if we don't like doing that

app/pages/SchemeBankAccountPage.scala
```diff
@@ -20,7 +20,7 @@ import models.BankAccount
 import models.SchemeId.Srn
 import play.api.libs.json.JsPath
 
-case class SchemeBankAccountPage(srn: Srn) extends QuestionPage[BankAccount] {
+case class SchemeBankAccountPage(srn: Srn) extends QuestionPage[List[BankAccount]] {
```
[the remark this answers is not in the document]
westwater: The only issue with this is how you would get all bank accounts from user answers as we won't have access to the list (other than traversing the user answers json obj manually / with separate reads)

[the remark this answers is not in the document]
westwater: oh you mentioned it here https://github.com/hmrc/pension-scheme-return-frontend/pull/18#discussion_r1109492196

app/navigation/Navigator.scala
westwater: this was more for readability buy can remove

app/config/Constants.scala
kurtlogan: If you do the refactor to the navigator you won't need this anymore

app/config/Binders.scala
westwater: I'm not sure if this is feasible - the Validate instance would look something like this
```
  implicit def max10Validator: Validate[Int, OneToTen] = Validate.instance(
    i => Result.fromBoolean(refineV[OneToTen](i).isRight, i),
    _.toString
  )
```
but since we use refineV (which needs a validator for OneToTen), it will use itself as the implicit instance and cause a StackOverflow - one way to get around it is to not use refine and do a boolean check on the int (i > 0 && <= 10) but thats not using the Refined which doesn't seem right

app/navigation/Navigator.scala
westwater: Spoke to design team and they said it should just go to the list page

kurtlogan (approved):

PR #22, opened 2023-02-21
PSR-102 startpage content change

narendrapaduchuri:


kurtlogan (dismissed):

westwater (approved):

app/views/ContentPageView.scala.html
```diff
@@ -34,7 +34,7 @@ <h1 class="govuk-heading-xl">@viewmodel.heading.toMessage</h1>
     <p class="govuk-body">
         @govukButton(
             ButtonViewModel(viewmodel.buttonText.toMessage)
-                    .asLink(viewmodel.onSubmit.url)
+                    .asLink(viewmodel.onSubmit.url).asStartButton()
```
kurtlogan, line 37: ContentPageView is a generic view for displaying content, not all content pages are start pages so setting the button as a start button will not always apply, you will need to configure that this is a start button in the view model and pass that to the button

kurtlogan (changes requested):

kurtlogan (approved):

westwater (approved):

PR #23, opened 2023-02-21
PSR-91 Added remove bank account page

westwater:


app/controllers/RemoveSchemeBankAccountController.scala
```diff
@@ -0,0 +1,97 @@
...
+            Future.successful(Redirect(navigator.nextPage(RemoveSchemeBankAccountPage(srn), mode, request.userAnswers)))
+          }
+        }
+      )
+  }
+
+  private def withBankAccountAtIndex(srn: Srn, index: Max10)(f: BankAccount => Result)(implicit request: DataRequest[_]): Result =
+    request.userAnswers.get(SchemeBankAccountPage(srn, index)) match {
+      case Some(bankAccount) => f(bankAccount)
+      case None => NotFound(s"Bank account at index ${index.value} not found")
```
kurtlogan, line 82: Not found won't work here, this will just display an empty page with the given text, if the bank account is not found that could be for a few reasons including a cache reaching its ttl, in these circumstances we should redirect to the journey recovery controller which will prompt the user on what to do

app/pages/RemoveSchemeBankAccountPage.scala
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package pages
+
+import play.api.libs.json.JsPath
+import models.SchemeId.Srn
+
+case class RemoveSchemeBankAccountPage(snr: Srn) extends QuestionPage[Boolean] {
```
kurtlogan, line 22:
```suggestion
case class RemoveSchemeBankAccountPage(srn: Srn) extends QuestionPage[Boolean] {
```

kurtlogan (changes requested):

kurtlogan (approved):

PR #41, opened 2023-03-13
PSR-114

westwater:


test/controllers/ControllerBehaviours.scala
```diff
@@ -123,7 +123,7 @@ trait ControllerBehaviours {
     redirectNextPage(call, defaultUserAnswers, form: _*)
 
   def redirectToPage(call: => Call, page: => Call, userAnswers: UserAnswers, form: (String, String)*): Unit =
-    s"redirect to page with form $form" in {
+    s"redirect to page ${page.url} with form $form" in {
```
kurtlogan, line 126: Although this might seem like a good idea it can break tests, the page value is passed in as a by name parameter because the play application wires up routes differently if an application is running or not so passing it in as a by name param means it will be evaluated in the body of the test when the application is running. By getting the url for the page to populate the test name this now evaluates the page value when no application is running potentially breaking tests

kurtlogan: One idea I had to get around the rigidness of the test names was to return a case class instead of Unit with a run method that would execute the test but could also optionally accept a test name.

Something like
```
case class BehaviourTest(name: String, test: () => Unit) {
  
  def run: Unit = run(name)

  def run(name: String): Unit =
    name in test()
}
```
usage 
```
def redirectToNextPage(call: => Call, page: => Call): BehaviourTest =
  BehaviourTest("Redirect to next page", () => {
     // test logic goes here
  })

behave like redirectToNextPage(call, nextPage).run
behave like redirectToNextPage(call, nextPage).run("Redirect to CYA page")
```

kurtlogan (approved):

kurtlogan (approved):

PR #48, opened 2023-03-16
PSR-118 Added why no NINO page

westwater:


kurtlogan (commented):
You will need to implement cleanup method for the `NationalInsuranceNumberPage` so if the user changes that answer then it will remove the unused value from the cache

app/pages/NationalInsuranceNumberPage.scala
```diff
@@ -19,10 +19,20 @@ package pages
...
+import scala.util.Try
 
 case class NationalInsuranceNumberPage(srn: Srn, index: Max99) extends QuestionPage[Boolean] {
 
   override def path: JsPath = JsPath \ toString
 
   override def toString: String = "nationalInsuranceNumber"
+
+  override def cleanup(value: Option[Boolean], userAnswers: UserAnswers): Try[UserAnswers] =
+    value.fold(Try(userAnswers))(if (_) {
```
kurtlogan, line 33: If the value is `None` it means the answer hs been removed so this case should remove both answers, something like 
```
value match {
  case Some(true)  => userAnswers.remove(NoNinoPage(srn, index))
  case Some(false) => userAnswers.remove(MemberDetailsNinoPage(srn, index))
  case None             => userAnswers.remove(NoNinoPage(srn, index))
                                     .flatMap(_.remove(MemberDetailsNinoPage(srn, index))
}
```

kurtlogan (changes requested):

app/pages/NationalInsuranceNumberPage.scala
kurtlogan: Also need some tests for this 😉

westwater: Added tests here https://github.com/hmrc/pension-scheme-return-frontend/blob/98331bf2747583df660961f6aed069355e02f4a3/test/pages/NationalInsuranceNumberPageSpec.scala#L35-L50 but good point around the None 👍

kurtlogan (approved):

PR #51, opened 2023-03-20
PSR-158 cache user selection

narendrapaduchuri:


app/controllers/PensionSchemeMembersController.scala
```diff
@@ -43,36 +43,53 @@ import pages.PensionSchemeMembersPage
...
+    extends FrontendBaseController
     with I18nSupport {
 
   private val form = PensionSchemeMembersController.form(formProvider)
 
   def onPageLoad(srn: Srn): Action[AnyContent] = identifyAndRequireData(srn) { implicit request =>
-    Ok(view(form, viewModel(srn, request.schemeDetails.schemeName)))
+    Ok(
+      view(
+        form.fromUserAnswers(PensionSchemeMembersPage(srn, Manual)),
```
westwater, line 74: Since we are saving the selection here I don't think we need to have `Manual` or `Upload` as part of the page case class - we can just pull it from useranswers in the navigator to chose where to go next 

It also makes this more readable as hardcoding Manual doesn't make sense

westwater (approved):

PR #52, opened 2023-03-21
Psr 98

kurtlogan:


app/models/UserAnswers.scala
```diff
@@ -50,6 +50,9 @@ final case class UserAnswers(
     }
   }
 
+  def unsafeSet[A](page: Settable[A], value: A)(implicit writes: Writes[A]): UserAnswers =
+    set(page, value).get
```
westwater, line 54: Not sure about defining unsafe methods outside of test scenarios

test-utils/utils/BaseSpec.scala
```diff
@@ -69,4 +71,45 @@ abstract class BaseSpec
     running(_ => applicationBuilder)(block)
 
   def urlEncode(input: String): String = URLEncoder.encode(input, "utf-8")
+
+  @deprecated("behave word has been replace with act word - behave.like becomes act.like")
+  override val behave: BehaveWord = new BehaveWord
```
westwater, line 76: Does this line need to be here?

westwater (approved):

test-utils/utils/BaseSpec.scala
kurtlogan: Yes so the only was I can discourage use of behave is to deprecate it, I originally wanted to throw an exception if it was used but behave in the super class is declared with a val so it must be a val in the subclass which doesn't work for exceptions, the only other thing I can do to let the developer know they have done something wrong is to deprecate the value

app/models/UserAnswers.scala
kurtlogan: Yeah me neither, I I'll go back and add it as an extension method

westwater (approved):

PR #62, opened 2023-03-27
PSR-129 trim all text field values

westwater:


test/forms/behaviours/FieldBehaviours.scala
```diff
@@ -103,6 +103,15 @@ trait FieldBehaviours extends FormSpec with ScalaCheckPropertyChecks with Genera
   def invalidField(form: Form[_], fieldName: String, errorMessage: String, invalidData: Gen[String], args: Any*): Unit =
     errorField("invalid field", form, fieldName, FormError(fieldName, errorMessage, args.toList), invalidData)
 
+  def trimmedField(form: Form[_], fieldName: String, gen: Gen[String]): Unit =
```
kurtlogan, line 106: Instead of adding this behaviour here why not change the `fieldThatBindsValidData` to also check that whitespace is trimmed. Something like:
```
  def fieldThatBindsValidData(form: Form[_], fieldName: String, validDataGenerator: Gen[String]): Unit =
    "bind valid data" in {

      def whitespace(gen: Gen[String]): Gen[(String, String)] =
        for {
          value <- gen
          ws1   <- Gen.listOf(" ").map(_.mkString)
          ws2   <- Gen.listOf(" ").map(_.mkString)
        } yield (
          (value, ws1 + value + ws2)
        )
      
      forAll(whitespace(validDataGenerator) -> "validDataItem") { case (original, whitespaced) =>
        val result = form.bind(Map(fieldName -> whitespaced)).apply(fieldName)
        result.value.value mustBe original
        result.errors mustBe empty
      }
    }
```
This will then be checked for all existing forms without needing to add another behaviour

kurtlogan (approved):

PR #106, opened 2023-05-02
PSR-151 Added upload member details page

westwater:


app/controllers/nonsipp/memberdetails/UploadMemberDetailsController.scala
```diff
@@ -0,0 +1,93 @@
...
+      "uploadMemberDetails.heading",
+      ParagraphMessage("uploadMemberDetails.paragraph"),
+      detailsContent = ParagraphMessage("uploadMemberDetails.details.paragraph") ++ ListMessage(
+        ListType.Bullet,
+        "uploadMemberDetails.list1",
+        "uploadMemberDetails.list2",
+        "uploadMemberDetails.list3"
+      ),
+      acceptedFileType = ".csv",
+      maxFileSize = "100MB",
```
westwater, line 88: these don't need to be in the messages file as it would be the same in welsh

app/viewmodels/DisplayMessage.scala
```diff
@@ -44,7 +44,7 @@ object DisplayMessage {
   case class CompoundMessage(message: DisplayMessage, next: DisplayMessage) extends InlineMessage
 
   implicit class DisplayMessageOps(message: DisplayMessage) {
-    def +(other: DisplayMessage): CompoundMessage = CompoundMessage(message, other)
+    def ++(other: DisplayMessage): CompoundMessage = CompoundMessage(message, other)
```
westwater, line 47: this was required as sometimes the compiler is trying to use the `+` from scala String to add messages together

narendrapaduchuri (dismissed):

narendrapaduchuri (approved):

PR #113, opened 2023-05-15
NO-JIRA Make upscan callback https + refactor upload logic

westwater:


app/repositories/UploadRepository.scala
```diff
@@ -53,15 +53,24 @@ class UploadRepository @Inject()(mongoComponent: MongoComponent)(implicit ec: Ex
   def find(key: UploadKey): Future[Option[UploadDetails]] =
     collection.find(equal("id", Codecs.toBson(key))).headOption()
 
-  def updateStatus(reference: Reference, newStatus: UploadStatus): Future[UploadStatus] =
+  def updateStatus(reference: Reference, newStatus: UploadStatus): Future[Unit] =
     collection
       .findOneAndUpdate(
         filter = equal("reference", Codecs.toBson(reference)),
```
kurtlogan, line 59: You can add additional codecs to `PlayMongoRepository` so you don't need to call `Codecs.toBson` on your types.

kurtlogan (changes requested):

app/repositories/UploadRepository.scala
westwater: Was having some trouble with this as some operations failed when I removed the Codecs.toBson even though I had Json format instances in scope

kurtlogan (approved):

PR #117, opened 2023-05-17
PSR-173 scheme hold any shares acquired

narendrapaduchuri:


kurtlogan (changes requested):

app/viewmodels/models/YesNoPageViewModel.scala
```diff
@@ -30,6 +30,8 @@ case class YesNoPageViewModel(
   yes: Option[Message] = None,
   no: Option[Message] = None,
   refresh: Option[Int] = None,
+  details: Option[DisplayMessage] = None,
```
kurtlogan, line 33: The details and display values are semantically connected as they are both needed to render the same element. Encoding it this was although it works is brittle as it requires _both_ values to be set but this is not enforced in code which could lead to errors.

For something like this you should create a new case class that encapsulates the properties then set the type on the `YesNoPageViewModel` to be optional.

Something like 
```
case class FurtherDetailsViewModel(
   title: Message,
   contents: DisplayMessage
)

case class YesNoPageViewModel(
  ...
  furtherDetails: Option[FurtherDetailsViewModel]
  ...
)
```

kurtlogan (dismissed):

PR #119, opened 2023-05-19
Update view models

kurtlogan:


app/viewmodels/models/PageViewModel.scala
```diff
@@ -0,0 +1,68 @@
...
+
+import play.api.mvc.Call
+import viewmodels.DisplayMessage
+import viewmodels.DisplayMessage.{InlineMessage, Message}
+
+case class PageViewModel[+A](
+  title: Message,
+  heading: InlineMessage,
+  description: Option[DisplayMessage],
+  inset: Option[DisplayMessage],
```
westwater, line 27: should description and inset be part of the underlying page model instead as they are optional? A case could be made for description as most pages have it but not sure about inset

PR #132, opened 2023-06-02
PSR-179 who received loan

narendrapaduchuri:


app/models/ReceivedLoanType.scala
```diff
@@ -0,0 +1,46 @@
...
+    case ReceivedLoanType.AUKCompany => aUKCompany
+    case ReceivedLoanType.AUKPartnership => aUKPartnership
+    case ReceivedLoanType.Other => other
+
+  }
+}
+
+object ReceivedLoanType extends Enumerable.Implicits {
+
+  case object AnIndividual extends WithName("anIndividual") with ReceivedLoanType
```
kurtlogan, line 36: Don't prefix these with `A` and `An` instead call them
`Individual`
`UKCompany`
`UKPartnership`
`Other`

app/viewmodels/models/RadioListViewModel.scala
```diff
@@ -64,4 +64,7 @@ object RadioListRowViewModel {
 
   def apply(content: Message, value: String, hint: Message): RadioListRowViewModel =
     RadioListRowViewModel(content, value, Some(hint))
+
+  def apply(content: Message): RadioListRowViewModel =
```
kurtlogan, line 68: You need a value for the radio list item, how will the selected answer ne matched to the corresponding ADT value?

kurtlogan (changes requested):

kurtlogan (approved):

PR #136, opened 2023-06-09
PSR-198 Added CYA page

westwater:


app/controllers/nonsipp/BasicDetailsCheckYourAnswersController.scala
```diff
@@ -0,0 +1,291 @@
...
+    request.minimalDetails.individualDetails match {
+      case Some(individual) => Right(individual.fullName)
+      case None =>
+        request.minimalDetails.organisationName match {
+          case Some(orgName) => Right(orgName)
+          case None => Left(Redirect(controllers.routes.JourneyRecoveryController.onPageLoad()))
+        }
+    }
+
+  private def requiredPage[A: Reads](page: Gettable[A])(implicit request: DataRequest[_]): Either[Result, A] =
```
kurtlogan, line 114: This would be useful for other controllers, think about putting it somewhere shared

kurtlogan (dismissed):

app/connectors/MinimalDetailsConnector.scala
```diff
@@ -46,6 +46,7 @@ class MinimalDetailsConnectorImpl @Inject()(appConfig: FrontendAppConfig, http:
   )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[MinimalDetailsError, MinimalDetails]] =
     fetch("pspId", pspId.value)
 
+  // API 1442 (Get psa/psp minimal details) - https://confluence.tools.tax.service.gov.uk/pages/viewpage.action?spaceKey=PODSP&title=PODS+API+Latest+Version
```
kberec, line 49: Is it ok to have confluence links in the code?

westwater: good shout! I think this is a public repo so i'll remove the links

app/controllers/nonsipp/BasicDetailsCheckYourAnswersController.scala
westwater: You did something similar but as an action but it only allows a single page to be required, i'll look at refactoring this out in the future and adapting your code to take multiple pages

PR #137, opened 2023-06-09
PSR-180 Added individual recipient name page

kberec:


kberec:
Two things can be still discussed for this and other tickets in the same journey:

1. Do we need refined numerics in the URLs similarly to member details and if so, do we want to do that with each ticket or separately at the end of the journey?
2. Do we want to move every controller/view/page of this journey into the loansmadeoroutstanding package?

kurtlogan (approved):

PR #142, opened 2023-06-13
fixed ttl for upload repo

kurtlogan:


app/repositories/SessionRepository.scala
```diff
@@ -50,7 +50,7 @@ class SessionRepository @Inject()(
             .expireAfter(appConfig.cacheTtl, TimeUnit.SECONDS)
         )
       ),
-      replaceIndexes = true
+      replaceIndexes = false
```
kberec, line 53: After lowering the ttl configuration to 60 seconds I've got an exception when running this locally: " 'Index with name: lastUpdatedIdx already exists with different options' on server localhost:27017. The full response is {"ok": 0.0, "errmsg": "Index with name: lastUpdatedIdx already exists with different options", "code": 85, "codeName": "IndexOptionsConflict"}"

kberec (approved):

PR #144, opened 2023-06-15
PSR-213 Added individual Nino page

westwater:


app/forms/TextFormProvider.scala
```diff
@@ -17,22 +17,22 @@
...
 
 import javax.inject.Inject
 
-class TextFormProvider @Inject()() extends Mappings {
+class TextFormProvider @Inject()() {
 
-  val nameRegex = "^[a-zA-Z\\-' ]+$"
-  val textAreaRegex = """^[a-zA-Z0-9\-'" \t\r\n,.@/]+$"""
-  val textAreaMaxLength = 160
+  protected[forms] val nameRegex = "^[a-zA-Z\\-' ]+$"
```
kberec, line 27: Do we need `[forms]` here? As it is already directly in forms package?

conf/messages.en
```diff
@@ -626,3 +626,23 @@ companyRecipientName.heading = What is the recipient’s company name?
 companyRecipientName.error.required = Enter the recipient’s company name
 companyRecipientName.error.invalid.characters = Recipient’s company name must only include letters, numbers, and special characters such as hyphens, spaces and apostrophes
 companyRecipientName.error.length = Recipient’s company name must be 160 characters or fewer
+
+isMemberOrConnectedParty.title = Is the recipient a member or a connected party?
```
kberec, line 630: Are these under `isMemberOrConnectedParty` used anywhere?

kberec (approved):

app/forms/TextFormProvider.scala
westwater: it's mainly so the test in the same package can access the regex to assert and also stops consuming classes from accessing them

conf/messages.en
westwater: I think there was an issues when I rebased as I was jumping between the 2 stories (this one and PSR-215) - I'm going to merge PSR-215 after this so I can sort out this issue then

PR #149, opened 2023-06-23
PSR-219 Added Upload too many rows page

westwater:


app/services/MemberDetailsUploadValidator.scala
```diff
@@ -164,9 +167,9 @@ class MemberDetailsUploadValidator @Inject()(
           case nameDOBFormProvider.firstName => ValidationErrorType.FirstName
           case nameDOBFormProvider.lastName => ValidationErrorType.LastName
           case nameDOBFormProvider.dateOfBirth => ValidationErrorType.DateOfBirth
-          case dobDayKey => ValidationErrorType.DateOfBirth
-          case dobMonthKey => ValidationErrorType.DateOfBirth
-          case dobYearKey => ValidationErrorType.DateOfBirth
+          case `dobDayKey` => ValidationErrorType.DateOfBirth
```
kberec, line 170: As a matter of interest, what is the reason for this change?

kberec (approved):

app/services/MemberDetailsUploadValidator.scala
westwater: it makes it pattern match against the value of `dobDayKey` defined above instead of assigning the value to `dobDayKey` to be used in the pattern matching function

kberec (approved):

PR #170, opened 2023-07-18
Psr 190 2

narendrapaduchuri:


app/views/ConditionalYesNoPageView.scala.html
```diff
@@ -93,6 +95,16 @@ <h1 class="govuk-heading-l">@renderMessage(viewmodel.heading)</h1>
     )
 }
 
+@security(key: String, message: DisplayMessage) = @{
```
westwater, line 98: I don't think we need this, we can just use the textArea to render this and add the hint

narendrapaduchuri: We can use textArea but I am not sure because we have added field type of security for it

westwater (dismissed):

PR #183, opened 2023-07-31
PSR-212 added financial CYA page and changed navigation to task list

narendrapaduchuri:


kberec (commented):
@narendrapaduchuri Would it be better for the new `financial details` classes you added to be in the schemeDesignatory package?

narendrapaduchuri:
> @narendrapaduchuri Would it be better for the new `financial details` classes you added to be in the schemeDesignatory package?

Hi @kberec, that would be better to move schemeDesignatory package, I moved it.

kberec (approved):

PR #184, opened 2023-08-04
PSR-271 finance details CYA change link

narendrapaduchuri:


app/models/Money.scala
```diff
@@ -19,7 +19,14 @@ package models
 import play.api.libs.json.{Format, Json}
 import utils.Transform
 
-case class Money(value: Double, displayAs: String)
+import java.text.DecimalFormat
+
+case class Money(value: Double, displayAs: String) {
+
+  def formattedDisplayAs(format: String): String =
```
kberec, line 26: Would it be possible to use this method every time when `displayAs` is referenced from the code? 
As discussed in slack https://hmrcdigital.slack.com/archives/C042149LQFN/p1691159814423129?thread_ts=1691153155.111939&cid=C042149LQFN. The goal would be to display it formatted everywhere it is used.

narendrapaduchuri: Hi @kberec , that makes sense, I will add it and push the latest code.

westwater (dismissed):

westwater (approved):

kberec (approved):

PR #194, opened 2023-08-09
PSR-266 - Spike generic approach for identity pages

kberec:
Spike for generic approach for identity pages


narendrapaduchuri (approved):

westwater:
PR looks good, should save us quite a bit of time - as long as we have a guarantee that these pages won't change shape between the flows then I'm happy with it

kberec:
> PR looks good, should save us quite a bit of time - as long as we have a guarantee that these pages won't change shape between the flows then I'm happy with it

Thanks @westwater, I done a demo to psr_design (FYI @joshhumphreys ) today and they were happy with it. We don't have the design for all the journeys yet, but we discussed all of them should follow the same pattern. We also discussed how URL path elements should contain a fixed part and 'subject' part and that was fine too.

westwater (approved):

PR #195, opened 2023-08-10
PSR-322 Amended loans routing

westwater:


app/navigation/nonsipp/LoansMadeOrOutstandingNavigator.scala
```diff
@@ -36,8 +38,14 @@ object LoansMadeOrOutstandingNavigator extends JourneyNavigator {
       }
 
     case WhatYouWillNeedLoansPage(srn) =>
-      controllers.nonsipp.loansmadeoroutstanding.routes.WhoReceivedLoanController
-        .onPageLoad(srn, refineMV(1), NormalMode)
+      // if final loans page was completed, route user to loans list page
```
kberec, line 41: I think we can have this as an addition way prevent the routing to point to the existing loan, but this be also sorted out (in a regular user flow) when the task-list statuses are introduced. I.e:

- if the loans are in completed status, task list will navigate to loan list page
- if the loans are not in completed status, task list will navigate to the starting point in the journey.

kberec (approved):

PR #198, opened 2023-08-14
PSR-206-1 why does scheme hold land property

narendrapaduchuri:


app/navigation/nonsipp/LandOrPropertyNavigator.scala
```diff
@@ -42,6 +42,9 @@ object LandOrPropertyNavigator extends JourneyNavigator {
 
     case LandOrPropertyAddressLookupPage(srn, index) =>
       controllers.routes.UnauthorisedController.onPageLoad()
+
+    case WhyDoesSchemeHoldLandPropertyPage(srn, index) =>
```
westwater, line 46: Is there any routing from a previous page to this page?

westwater (dismissed):
LGTM just a small comment

cyeniceri (dismissed):

app/navigation/nonsipp/LandOrPropertyNavigator.scala
cyeniceri: this should call the page (my ticket) is not ready yet

narendrapaduchuri: Hi George, Yes there is a routing from previous page, but that page is in block now(PSR-205).

narendrapaduchuri: @westwater , I think it's better to set navigation from address lookup page to this page as of now. Otherwise page is not displaying. Once we get the other page in between we can change the routing accordingly.

westwater (approved):

PR #215, opened 2023-08-18
PSR-302 Error message changed according to the new regex

AlessandroCinque:
PSR-302 as the Doc/ticket requested https://docs.google.com/document/d/1VKy1RMKgFk8jLMLflygoWaPBBHp-t4GWKWFUIVxRImI/edit the error has been changed.


cyeniceri:
do we need to change regExp as well along with the message ?

AlessandroCinque:
> I didn't see regexp change, is that normal ?

No need for that as it is taking it from the Yes No controller 
image

cyeniceri (approved):
lgtm

PR #225, opened 2023-08-29
PSR-267 Added min values to money formatter

westwater:


kberec:
As of 6/10/2023 we are waiting for a confirmation if this need to be applied to any fields. Reference: comment thread of the ticket.

PR #236, opened 2023-09-01
PSR-249

kberec:


AlessandroCinque (approved):
Seems all good at first sight for me, though I must admit that I still do not fully understand the functioning of the IdentityType, therefore I recommend  for the review of another dev too

westwater (approved):

PR #239, opened 2023-09-05
PSR-230 & 274  Is Lessee Connected Party

narendrapaduchuri:


app/pages/nonsipp/landorproperty/IsLesseeConnectedPartyPage.scala
```diff
@@ -0,0 +1,29 @@
...
+import config.Refined.Max5000
+import models.SchemeId.Srn
+import pages.QuestionPage
+import play.api.libs.json.JsPath
+
+case class IsLesseeConnectedPartyPage(srn: Srn, index: Max5000) extends QuestionPage[Boolean] {
+
+  override def path: JsPath = JsPath \ toString
+
+  override def toString: String = "isLesseeConnectedParty"
```
kberec, line 28: Need to be a path in land or property structure. Could you please also update the mapping table?

narendrapaduchuri: Sorry forgot to add that,  I have just updated both

kberec: Thanks for changing it @narendrapaduchuri.... would it be ok to use the constants defined instead of hardcoded values please, as for example in WhyDoesSchemeHoldLandPropertyPage?  This is more for a future reference than any change for this functionality.

PR #242, opened 2023-09-05
PSR-327

westwater:
Branched from PSR-209

### DIFF
--- a/app/pages/nonsipp/landorpropertydisposal/LandOrPropertyStillHeldPage.scala
+++ b/app/pages/nonsipp/landorpropertydisposal/LandOrPropertyStillHeldPage.scala
@@ -14,14 +14,21 @@
  * limitations under the License.
  */
 
-package pages.nonsipp
+package pages.nonsipp.landorpropertydisposal
 
-import play.api.libs.json.{__, JsPath}
+import play.api.libs.json.JsPath
+import models.SchemeId.Srn
+import models.Money
+import pages.QuestionPage
+import config.Refined.{Max50, Max5000}
+import utils.RefinedUtils._
+import eu.timepit.refined.refineMV
 
-package object landorpropertydisposal {
-  object Paths {
-    val assets: JsPath = __ \ "assets"
-    val landOrProperty: JsPath = assets \ "landOrProperty"
-    val disposalPropertyTransaction: JsPath = landOrProperty \ "heldPropertyTransaction" \ "disposedPropertyTransaction"
-  }
+case class LandOrPropertyStillHeldPage(srn: Srn, landOrPropertyIndex: Max5000, disposalIndex: Max50)
+    extends QuestionPage[Boolean] {
+
+  override def path: JsPath =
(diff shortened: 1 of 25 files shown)


app/pages/nonsipp/landorpropertydisposal/LandOrPropertyStillHeldPage.scala
```diff
@@ -0,0 +1,32 @@
...
+import pages.QuestionPage
+import config.Refined.Max5000
+import utils.RefinedUtils._
+import eu.timepit.refined.refineMV
+
+case class LandOrPropertyStillHeldPage(srn: Srn, index: Max5000) extends QuestionPage[Boolean] {
+
+  override def path: JsPath = JsPath \ toString \ index.arrayIndex.toString
+
+  override def toString: String = "landOrPropertyStillHeld"
```
kberec, line 31: This is a tricky one. It's the first one where there is an array in an array. Data path is `assets.landOrProperty.landOrPropertyTransactions[].heldPropertyTransaction.disposedPropertyTransaction[n].portionStillHeld`  So multiple disposals can be created for the same heldPropertyTransaction.

conf/landorpropertydisposal.routes
```diff
@@ -2,3 +2,7 @@ GET        /:srn/land-property-disposal/:index                 controllers.nonsi
 POST       /:srn/land-property-disposal/:index                 controllers.nonsipp.landorpropertydisposal.LandOrPropertyDisposalController.onSubmit(srn: Srn, index: Max5000, mode: Mode = NormalMode)
 GET        /:srn/change-land-property-disposal/:index          controllers.nonsipp.landorpropertydisposal.LandOrPropertyDisposalController.onPageLoad(srn: Srn, index: Max5000, mode: Mode = CheckMode)
 POST       /:srn/change-land-property-disposal/:index          controllers.nonsipp.landorpropertydisposal.LandOrPropertyDisposalController.onSubmit(srn: Srn, index: Max5000, mode: Mode = CheckMode)
+GET        /:srn/any-part-land-property-still-held/:index                 controllers.nonsipp.landorpropertydisposal.LandOrPropertyStillHeldController.onPageLoad(srn: Srn, index: Max5000, mode: Mode = NormalMode)
```
kberec, line 5: Multiple disposals can be created for the same heldPropertyTransaction. Not sure if there is any need to reflect that in the URL, just highlighting it anyway, just for consideration.

westwater: added a double index in the url, can't think of a cleaner solution as we need to propagate both positions when navigating between pages

app/pages/nonsipp/landorpropertydisposal/LandOrPropertyStillHeldPage.scala
westwater: changed the path to a double index to capture this

kberec (approved):

PR #281, opened 2023-09-27
PSR-420 failed tests are fixed and additional transformation tests ar…

cyeniceri:
…e added, utr bug is fixed


app/models/Utr.scala
```diff
@@ -33,7 +33,7 @@ object Utr extends (String => Utr) {
   implicit val utrWrite: Writes[Utr] = new SimpleObjectWrites[Utr](_.value)
   implicit val utrRead: Reads[Utr] = new SimpleObjectReads[Utr]("utr", Utr.apply)
 
-  private val validUtrFormat = """^\d{5}\s*?\d{5}"""
+  private val validUtrFormat = """^[0-9]{10}$"""
```
kberec, line 36: I think we should keep the validation here as it was, as it was designed, i.e. to allow with or without a space here.
We can trim out spaces in the transformation when sending to ETMP.
Unless you discussed it with PSR design team and they agreed to change the design for this.

build.sbt
```diff
@@ -47,10 +48,10 @@ lazy val root = (project in file("."))
     PlayKeys.playDefaultPort := 10701,
     ScoverageKeys.coverageExcludedFiles := "<empty>;Reverse.*;.*handlers.*;.*components.*;" +
       ".*Routes.*;.*viewmodels.govuk.*;",
-    ScoverageKeys.coverageMinimumStmtTotal := 78,
+    ScoverageKeys.coverageMinimumStmtTotal := 83,
```
kberec, line 51: Not necessarily now, but gradually we'd need to go back to 90 here

app/services/PSRSubmissionService.scala
```diff
@@ -17,219 +17,44 @@
...
+  psrConnector: PSRConnector,
+  minimalRequiredSubmissionTransformer: MinimalRequiredSubmissionTransformer,
+  loanTransactionsTransformer: LoanTransactionsTransformer
+) {
 
   def submitMinimalRequiredDetails(
     srn: Srn
   )(implicit hc: HeaderCarrier, ec: ExecutionContext, request: DataRequest[_]): Future[Option[Unit]] =
-    buildMinimalRequiredSubmission(srn).map(psrConnector.submitMinimalRequiredDetails(_)).sequence
+    minimalRequiredSubmissionTransformer.transform(srn).map(psrConnector.submitMinimalRequiredDetails(_)).sequence
```
kberec, line 41: We would also need to remove psrConnector.submitMinimalRequiredDetails(_) from here and the BE and use  psrConnector.submitPsrDetails instead. Not sure if this is the right moment to do that or do it as a part of other ticket or create a bug. What is your opinion @cyeniceri ?

app/models/Utr.scala
cyeniceri: We are still allowing the users enter with/without space but we are removing whitespace(s) before executing the validation (line:23). Is this what you mean @kberec ? 

https://design.tax.service.gov.uk/hmrc-design-patterns/unique-taxpayer-reference/
(Remove spaces, characters and extra numbers before validating.)

build.sbt
cyeniceri: I made it 78 back 👍

kberec: That's fine, we can monitor if for a while in the pipeline and we can change it back later to 90 when we are confident we are continuously above 90.

app/services/PSRSubmissionService.scala
cyeniceri: submitMinimalRequiredDetails and related codes are removed. We are only using submitPsrDetails for all submissions including minimalRequiredSubmission

kberec (approved):

PR #296, opened 2023-10-10
BDOG-2783 configure play.http.router to use prod.Routes to follow pla…

Shnick:
…tform guidelines

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -16,7 +16,7 @@ include "frontend.conf"
 
 appName="pension-scheme-return-frontend"
 
-play.http.router = testOnlyDoNotUseInAppConf.Routes
+play.http.router = prod.Routes
 
 play.filters.enabled += play.filters.csp.CSPFilter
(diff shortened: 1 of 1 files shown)


conf/application.conf
```diff
@@ -16,7 +16,7 @@ include "frontend.conf"
 
 appName="pension-scheme-return-frontend"
 
-play.http.router = testOnlyDoNotUseInAppConf.Routes
+play.http.router = prod.Routes
```
kberec, line 19: We would like to keep using `test-only` routes defined in testOnlyDoNotUseInAppConf.routes. How should we do that? Would it be the opposite way around, i.e. adding testOnlyDoNotUseInAppConf.routes to prod.routes? FYI @westwater @cyeniceri

Shnick: Hey I'll explain the change in your team channel - there's connecting app config changes too.

westwater (approved):

PR #306, opened 2023-10-16
PSR-415 Added initial audit event

westwater:


conf/application.conf
```diff
@@ -97,6 +98,16 @@ microservice {
         secure = false
         maxFileSize = 50
       }
+
+      auditing {
+        enabled = false
+        consumer {
+          baseUri {
+            host = localhost
```
kberec, line 106: Do we need a config change for this for the cloud environments?

kberec (approved):

PR #311, opened 2023-10-18
PSR-494 Updated task list status for disposals

westwater:


app/controllers/nonsipp/TaskListController.scala
```diff
@@ -456,7 +459,7 @@ object TaskListController {
             .onPageLoad(srn, NormalMode)
```
kberec, line 459: Is the navigation the same for all scenarios?

westwater: It's not specified as AC but I can route the user straight to the disposal list page when at least 1 disposal is completed - don't know if that would be another story / if we have that on the backlog somewhere or if we do it as part of this

kberec: I was thinking about the last line of each AC: "and link navigates to...."

westwater: woops missed that, will change

kberec (dismissed):

kberec (approved):

PR #350, opened 2023-11-07
PSr-476 CIP Explicit event on validation

kberec:


narendrapaduchuri (approved):

app/controllers/nonsipp/memberdetails/CheckMemberDetailsFileController.scala
```diff
@@ -158,14 +171,44 @@ class CheckMemberDetailsFileController @Inject()(
     duration
   )
 
-  private def auditDownload(srn: Srn, responseStatus: Int, startTime: Long)(implicit request: DataRequest[_]): Unit = {
-    val endTime = System.currentTimeMillis
-    val duration = endTime - startTime
+  private def auditDownload(srn: Srn, responseStatus: Int, duration: Long)(implicit request: DataRequest[_]): Unit =
```
cyeniceri, line 174: it seems auditDownload/auditValidation method signature says return type unit but we are returning something else. It's not a big deal but a bit confusing

app/controllers/nonsipp/memberdetails/CheckMemberDetailsFileController.scala
```diff
@@ -158,14 +171,44 @@ class CheckMemberDetailsFileController @Inject()(
...
+    for {
+      taxYear <- schemeDateService.taxYearOrAccountingPeriods(srn).merge.getOrRecoverJourney
+      _ = auditService.sendEvent(buildValidationAuditEvent(taxYear, outcome))
+    } yield taxYear
+
+  private def buildValidationAuditEvent(taxYear: DateRange, outcome: (Upload, Int, Long))(
+    implicit req: DataRequest[_]
+  ) = PSRFileValidationAuditEvent(
+    schemeName = req.schemeDetails.schemeName,
+    schemeAdministratorName = req.schemeDetails.establishers.headOption.get.name,
```
cyeniceri, line 192: headOption.get doesn't make sense to me, why don't we use head directly ?

cyeniceri (approved):

app/controllers/nonsipp/memberdetails/CheckMemberDetailsFileController.scala
kberec: Will push the changes as a separate branch @cyeniceri as this was merged.

cyeniceri: thanks @kberec

PR #353, opened 2023-11-08
PSR-531 Fix for page navigation

AlessandroCinque:
https://jira.tools.tax.service.gov.uk/browse/PSR-531 Fixing the page navigation issue in the Land or Property disposal for the address list + some tiding up


AlessandroCinque:
image
Tests are happy

PR #365, opened 2023-11-14
PSR-396 Fixing merging errors

AlessandroCinque:
I noticed that there were some errors when I was resolving the conflicts for https://github.com/hmrc/pension-scheme-return-frontend/pull/361 , so I made this PR to fix them


AlessandroCinque:
To be more specific while renaming my controller in https://github.com/hmrc/pension-scheme-return-frontend/pull/361 I deleted by mistake some messages that were from another "What You Will" page thinking they were mine. To avoid confusion I did some further refactoring to make them more readable and distinguishable.

narendrapaduchuri (approved):

PR #367, opened 2023-11-16
PSR-480 Make all tests run in parallel

westwater:


kberec:
FYI @westwater I'm closing this as it become stale.

PR #381, opened 2023-11-24
PSR-627 library replaced with https://github.com/alphagov/accessible-…

cyeniceri:
…autocomplete because of jquery vulnerability in location-autocomplete.min.js


AlessandroCinque (approved):
LGTM

PR #396, opened 2023-12-01
PSR-557 Pension Scheme Return - feedback_frontend fix

cyeniceri:


app/views/templates/Layout.scala.html
```diff
@@ -45,6 +45,7 @@
                 keepAliveUrl        = Some(routes.KeepAliveController.keepAlive().url),
                 keepAliveButtonText = Some(messages("timeout.keepAlive")),
                 signOutUrl          = Some(controllers.auth.routes.AuthController.signOut().url),
+                timeoutUrl          = Some(controllers.auth.routes.AuthController.signOutNoSurvey().url),
```
kberec, line 48: It is fine, when I click sign out manually. But if the timer comes up and signs me out, still redirects me to the wrong URL:`http://localhost:9025/gg/sign-out?continue=%2Fpension-scheme-return%2Faccount%2Fsigned-out`

PR #401, opened 2023-12-05
PSR 481 remove address from list

AlessandroCinque:
https://jira.tools.tax.service.gov.uk/browse/PSR-481 If the user says that no portion of the properties are held, this will make sure the address doesn't pop in the list


AlessandroCinque:
image
 tests are happy

cyeniceri (approved):

PR #402, opened 2023-12-05
PSR-392 Remove Member Contribution

AlessandroCinque:
https://jira.tools.tax.service.gov.uk/browse/PSR-392 Remove page implemented and view optimised


AlessandroCinque:
image
tests are happy

app/controllers/nonsipp/membercontributions/RemoveMemberContributionController.scala
```diff
@@ -0,0 +1,136 @@
...
+                  RemoveMemberContributionController.viewModel(srn, memberIndex, index, total, nameDOB.fullName)
+                )
+              )
+            ).merge
+          },
+          removeDetails => {
+            if (removeDetails) {
+              for {
+                updatedAnswers <- Future
+                  .fromTry(request.userAnswers.remove(TotalMemberContributionPage(srn, memberIndex, index)))
```
kberec, line 99: page 19, MemberContributionsPage (which is in the wrong package btw) should be removed too if we are removing the last contribution. If it remains, it will be a problem when posting data into ETMP. For reference please see: EmployerNamePage

AlessandroCinque: I am moving MemberContributionsPage into the `membercontributions` package

kberec (approved):

PR #417, opened 2023-12-11
PSR-574 post transformation of landOrProperty disposal data

cyeniceri:


app/services/PsrSubmissionService.scala
```diff
@@ -60,11 +62,13 @@ class PsrSubmissionService @Inject()(
           minimalRequiredSubmission = minimalRequiredSubmission,
           checkReturnDates = checkReturnDates,
           loans = Option.when(schemeHadLoans)(Loans(schemeHadLoans, loanTransactionsTransformer.transformToEtmp(srn))),
-          assets = Option.when(landOrPropertyHeld || moneyWasBorrowed)(
+          assets = Option.when(landOrPropertyHeld || moneyWasBorrowed || disposeAnyLandOrProperty)(
```
kberec, line 65: What happens in a scenario when there are properties (or borrowings) but no disposals?

cyeniceri:
```
 optDisposedPropertyTransaction =
                      Option.when(disposeAnyLandOrProperty)(buildOptDisposedPropertyTransactions(srn, index))
```

Then optDisposedPropertyTransaction will be None and we'll send nothing to BE

kberec: Is that right though? We should send landOrProperty (and borrowing) even if there are no disposals.

cyeniceri: yes I mean, we are not going to send any DisposedPropertyTransaction (if false) but landOrProperty (and borrowing).

kberec (approved):

PR #425, opened 2023-12-14
PSR-544 | RemovePcls implementation

ccatosdevelopment:


app/pages/nonsipp/memberreceivedpcls/package.scala
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package pages.nonsipp
+
+import play.api.libs.json.{__, JsPath}
+
+package object memberreceivedpcls {
```
kberec, line 21: Good shout about adding this, in theory the same should be used in `PensionCommencementLumpSumAmountPage`. But we can do that later, no need to commit again.

kberec (dismissed):

kberec (approved):

PR #434, opened 2023-12-27
PSR-682 Added POST for transfer in

westwater:


app/controllers/nonsipp/receivetransfer/ReportAnotherTransferInController.scala
```diff
@@ -80,9 +81,13 @@ class ReportAnotherTransferInController @Inject()(
                   .set(ReportAnotherTransferInPage(srn, index, secondaryIndex), value)
               )
               _ <- saveService.save(updatedAnswers)
-            } yield Redirect(
-              navigator
-                .nextPage(ReportAnotherTransferInPage(srn, index, secondaryIndex), mode, updatedAnswers)
+              submissionResult <- psrSubmissionService.submitPsrDetails(srn, updatedAnswers)
```
kberec, line 84: Is there a reason why to submit PSR details here as well and not only on the submit of CYA page?

westwater: If they add additional transfers and don't go to the cya page none of them will be submitted, are we happy to do that?

kberec: I think the initial idea was to do a submission from the following places (from AC):
`Given that a PSR submission is made (on CYA or remove or inital yes/no or member list page)`
I'm not saying it is wrong to submit here too.
I'd suggest to keep it for now, but get a clarification from a wider group.

kberec (approved):

PR #454, opened 2024-01-10
PSR-559 | SurrenderedBenefitsAmount

ccatosdevelopment:


app/config/Constants.scala
```diff
@@ -58,6 +58,8 @@ object Constants {
   val minMoneyValue = -999999999.99
 
   val unallocatedContributionMin = 0.01
+  val minInputAmount = 0.01
+  val maxInputAmount = 999999999.99
```
kberec, line 62: I suggest the naming to be specific to surrender benefits (or use a generic constant, if there is one). Can be done in separate ticket too, if that spares some time.

kberec (approved):

kberec (approved):

PR #458, opened 2024-01-12
PSR-554 remove pension payments page

narendrapaduchuri:


app/controllers/nonsipp/memberpensionpayments/RemovePensionPaymentsController.scala
```diff
@@ -0,0 +1,146 @@
...
+)(implicit ec: ExecutionContext)
+    extends PSRController
+    with I18nSupport {
+
+  private val form = RemovePensionPaymentsController.form(formProvider)
+
+  def onPageLoad(srn: Srn, index: Max300): Action[AnyContent] =
+    identifyAndRequireData(srn) { implicit request =>
+      val nameDOB = request.userAnswers.get(MemberDetailsPage(srn, index)).get
+      val totalContrib = request.userAnswers.get(TotalAmountPensionPaymentsPage(srn, index))
```
cyeniceri, line 58: naming seems not correct here

app/navigation/nonsipp/PensionPaymentsReceivedNavigator.scala
```diff
@@ -16,13 +16,15 @@
 
 package navigation.nonsipp
 
+import controllers.nonsipp.memberpensionpayments.RemovePensionPaymentsController
```
cyeniceri, line 19: this import is not used here

app/pages/nonsipp/memberpensionpayments/RemovePensionPaymentsPage.scala
```diff
@@ -0,0 +1,26 @@
...
+
+package pages.nonsipp.memberpensionpayments
+
+import config.Refined.Max300
+import models.SchemeId.Srn
+import pages.Page
+
+case class RemovePensionPaymentsPage(srn: Srn, index: Max300) extends Page {
+
+  override def toString: String = "removePensionPayments"
```
cyeniceri, line 25: do we need this ? Are we using "removePensionPayments" anywhere such as in our userAnswer ?

cyeniceri (changes requested):

app/pages/nonsipp/memberpensionpayments/RemovePensionPaymentsPage.scala
narendrapaduchuri: It seems we don't need this, just checking now after that will push the code. Thanks

cyeniceri (approved):

PR #459, opened 2024-01-12
PSR-571

B-Savage:


app/pages/nonsipp/shares/CompanyNameRelatedSharesPage.scala
```diff
@@ -0,0 +1,29 @@
...
+import config.Refined.Max5000
+import models.SchemeId.Srn
+import pages.QuestionPage
+import play.api.libs.json.JsPath
+
+case class CompanyNameRelatedSharesPage(srn: Srn, index: Max5000) extends QuestionPage[String] {
+
+  override def path: JsPath = Paths.shares \ toString
+
+  override def toString: String = "companyNameRelatedShares"
```
kberec, line 28: We should have this as `shareIdentification.nameOfSharesCompany`, I suggest we get on a call and I can explain the reasons why

kberec (approved):

PR #460, opened 2024-01-12
Updated 3 employer contributions pages

MaryCong:
- What was the total contribution made by employer x
- Were employer contributions made by other employers 
- Check your answers and change details (employer contributions)


joshhumphreys (commented):
One outstanding question RE row 2028 - looks as though it's been deleted and replaced with the same thing, will wait for confirmation from @MaryCong before I approve.

kberec:
> One outstanding question RE row 2028 - looks as though it's been deleted and replaced with the same thing, will wait for confirmation from @MaryCong before I approve.

I think that's just the new line character removed.

MaryCong:
Yeah, I don't remember touching that line, so not sure what happened there. @kberec, if it's not going to mess up the code, is it ok to be approved?

kberec:
> Yeah, I don't remember touching that line, so not sure what happened there. @kberec, if it's not going to mess up the code, is it ok to be approved?

There are some conflicts we need to resolve first. We can do it on Monday.

MaryCong:
Sure, no rush, thanks!

kberec (approved):

PR #462, opened 2024-01-15
Update messages.en

joshhumphreys:
Updating '1' to 'one'
Updating 'active' bank account to 'open' bank account


MaryCong:
Looks good to me :) Don't think you made the change in row 2102 right? It's similar to row 2028 in my PR.

joshhumphreys:
> Looks good to me :) Don't think you made the change in row 2102 right? It's similar to row 2028 in my PR.

No, so not sure why this line us showing as changed! Can discuss in our 4.15 call

conf/messages.en
```diff
@@ -362,9 +362,9 @@ removeMemberDetails.title = Are you sure you want to remove this member?
 removeMemberDetails.heading = Are you sure you want to remove {0}?
 removeMemberDetails.error.required = Select yes if you are sure you want to remove
 
-activeAccountDetails.title = Did the scheme have at least 1 active bank account during the period of this return?
```
kberec, line 365: Unfortunately, QA's acceptance tests will fail if we change titles.
So we would need to ask QAs to change their tests too. It's a similarly simple change on their end, but it will fail if we just merge this in.

BabuRupoll7623: All fixed now

kberec: Thanks @BabuRupoll7623 , In that case, @joshhumphreys please resolve the conflicts and I'll merge this

BabuRupoll7623 (approved):

PR #464, opened 2024-01-15
PSR-652 Changed description to inset

kberec:


app/controllers/nonsipp/employercontributions/EmployerContributionsMemberListController.scala
```diff
@@ -227,13 +226,18 @@ object EmployerContributionsMemberListController {
     FormPageViewModel(
       title = Message(title, memberList.size),
       heading = Message(heading, memberList.size),
-      description = Some(ParagraphMessage("employerContributions.MemberList.paragraph")),
+      description = None,
       page = ActionTableViewModel(
-        inset = "employerContributions.MemberList.inset",
+        inset = ParagraphMessage(
```
cyeniceri, line 231: Are we sure that we are going to show "You must tell us about ......" always ??
I'm asking this because some of the journey we have fixed in this sprint, they are conditional. If you reach the maximum, we were hiding this and showing an warning message instead.

FYI @harishmuddegowda

app/controllers/nonsipp/employercontributions/EmployerContributionsMemberListController.scala
```diff
@@ -227,13 +226,18 @@ object EmployerContributionsMemberListController {
...
+          "employerContributions.MemberList.paragraph1"
+        ) ++
+          ParagraphMessage(
+            "employerContributions.MemberList.paragraph2"
+          ),
         head = Some(List(TableElem("Member name"), TableElem("Status"))),
         rows = rows(srn, mode, memberList, userAnswers),
         radioText = Message("employerContributions.MemberList.radios"),
-        showRadios = memberList.length < maxNotRelevant,
+        showInsetWithRadios = true,
```
cyeniceri, line 240: is it always true and no limit to add a new one ?

cyeniceri (commented):
My 2 comment is same for all journeys.
@harishmuddegowda 
https://jira.tools.tax.service.gov.uk/browse/PSR-652

app/controllers/nonsipp/employercontributions/EmployerContributionsMemberListController.scala
kberec: Yes, here it is always true, as the question is "Have you finished....?"

app/controllers/nonsipp/employercontributions/EmployerContributionsMemberListController.scala
kberec: Yes, I'm pretty sure. For member payments section this should be always visible.
For other journeys there is some optional inset text, which is shown when some kind of maximum is reached.

cyeniceri (approved):

PR #469, opened 2024-01-16
Updated transfer in and transfers out

MaryCong:
- Do you need to report another transfer in for [name of member]?
- Do you need to report another transfer out for [name of member]?


joshhumphreys (approved):
Approved

PR #506, opened 2024-01-31
PSR-742 Zap pipeline failure -- Absence of Anti-CSRF Tokens

cyeniceri:


conf/app.routes
```diff
@@ -16,8 +16,6 @@
 
 GET         /                                                                           controllers.IndexController.onPageLoad()
 
-# upload file notification callback
-#NOCSRF
```
kberec, line 20: @westwater if you think this is needed please let us know and we can have a chat about it

kberec (approved):

PR #507, opened 2024-01-31
PSR-602 shares total income page

narendrapaduchuri:


app/navigation/nonsipp/SharesNavigator.scala
```diff
@@ -161,7 +161,21 @@ object SharesNavigator extends JourneyNavigator {
       controllers.nonsipp.shares.routes.SharesIndependentValuationController.onPageLoad(srn, index, NormalMode)
 
     case SharesIndependentValuationPage(srn, index) =>
+      userAnswers.get(WhyDoesSchemeHoldSharesPage(srn, index)) match {
+        case Some(Acquisition) =>
+          controllers.routes.UnauthorisedController.onPageLoad()
```
ccatosdevelopment, line 166: If this is for the conditional page, then within this `Acquisition` branch I think there also needs to be a check that `TypeOfShares` is `SponsoringEmployer`. And if it's `Unquoted` or `ConnectedParty`, then it navigates `SharesTotalIncomeController` instead.

app/navigation/nonsipp/SharesNavigator.scala
```diff
@@ -161,7 +161,21 @@ object SharesNavigator extends JourneyNavigator {
...
+        case _ =>
```
ccatosdevelopment, line 167: If it's a `Contribution` or `Transfer`, then I think the navigation should go to `SharesTotalIncomeController` in any case, regardless of whether `TypeOfShares` is `SponsoringEmployer`, `Unquoted`, or `ConnectedParty`.

app/navigation/nonsipp/SharesNavigator.scala
BabuRupoll7623: changing now

app/navigation/nonsipp/SharesNavigator.scala
BabuRupoll7623: we are discussing same here, changing now

ccatosdevelopment (approved):

PR #524, opened 2024-02-08
PSR-856 Added return history page

kberec:


app/views/components/ReturnHistoryTableView.scala.html
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import viewmodels.ReturnHistorySummary
+
+@(rows: Seq[ReturnHistorySummary], colOne: String, colTwo: String, colThree: String, colFour: String, colFive: String, caption: String)(implicit messages: Messages)
```
westwater, line 19: very minor: as this is a component the class name should probably just be ReturnHistoryTable

Also I take it there was a reason raw html was chosen over the GovUkTable component?

westwater (approved):

westwater (approved):

app/views/components/ReturnHistoryTableView.scala.html
kberec: Renamed components. Reason for raw html is that I based this on other project's component. This is temporary solution for now, if it stays we can refactor.

PR #555, opened 2024-02-20
Psr 764

narendrapaduchuri:


app/pages/nonsipp/shares/WhyDoesSchemeHoldSharesPage.scala
```diff
@@ -45,5 +46,9 @@ case class WhyDoesSchemeHoldSharesPage(srn: Srn, index: Max5000) extends Questio
     }
 
   private def dependantPages(srn: Srn): List[Removable[_]] =
-    List(WhenDidSchemeAcquireSharesPage(srn, index))
+    List(
```
cyeniceri, line 49: Do you need to update WhyDoesSchemeHoldSharesPageSpec after this change ?

app/navigation/nonsipp/SharesNavigator.scala
```diff
@@ -213,120 +213,203 @@ object SharesNavigator extends JourneyNavigator {
...
     userAnswers => {
 
+      case TypeOfSharesHeldPage(srn, index) =>
+        userAnswers.get(TypeOfSharesHeldPage(srn, index)) match {
+          case Some(SponsoringEmployer) =>
+            controllers.nonsipp.shares.routes.WhyDoesSchemeHoldSharesController.onPageLoad(srn, index, CheckMode)
+          case Some(Unquoted) =>
+            controllers.nonsipp.shares.routes.WhyDoesSchemeHoldSharesController.onPageLoad(srn, index, CheckMode)
+          case Some(ConnectedParty) =>
+            controllers.nonsipp.shares.routes.SharesCYAController.onPageLoad(srn, index, CheckMode)
```
cyeniceri, line 223: For all these navigation changes in check routes, do we need to add corresponding unit tests in SharesNavigatorSpec ?

app/pages/nonsipp/shares/WhyDoesSchemeHoldSharesPage.scala
narendrapaduchuri: Hi @cyeniceri I believe we are not updating the this WhyDoesSchemeHoldSharesPageSpec.

app/navigation/nonsipp/SharesNavigator.scala
narendrapaduchuri: Yes, we need to fix @cyeniceri. Let me check if I can fix in this or next one

app/pages/nonsipp/shares/WhyDoesSchemeHoldSharesPage.scala
kberec: I think github shows this change with a mistake. I did this change (and updated WhyDoesSchemeHoldSharesPageSpec too) and that change has gone into main already.
@narendrapaduchuri could you please rebase on main? That might remove this change for the PR.

cyeniceri (approved):

PR #562, opened 2024-02-22
PSR-820 BE Not sending any information to ETMP regarding the No optio…

cyeniceri:
…n when use selects No option on initial page of all the journey's


app/models/requests/psr/Assets.scala
```diff
@@ -21,7 +21,7 @@ import play.api.libs.json._
 
 import java.time.LocalDate
 
-case class Assets(landOrProperty: LandOrProperty, borrowing: Borrowing)
+case class Assets(optLandOrProperty: Option[LandOrProperty], optBorrowing: Option[Borrowing])
```
kberec, line 24: Just to double check if similar change is needed on loans. I.e. something like:

```
case class Loans(schemeHadLoans: Boolean, loanTransactions: Option[Seq[LoanTransactions]])
```

kberec (approved):

kberec (approved):

app/models/requests/psr/Assets.scala
cyeniceri: https://github.com/hmrc/pension-scheme-return/pull/55/commits/00795f43d1a3638e3e0faf5896b2def7a3a9b5c1#diff-31e8e128877488c6f582653c425a1e2c285d2d5147cda7c4a069eaffca0a49c7R27

PR #572, opened 2024-02-29
PSR-829 What is other asset

B-Savage:


app/navigation/nonsipp/OtherAssetsHeldNavigator.scala
```diff
@@ -16,6 +16,8 @@
 
 package navigation.nonsipp
 
+import cats.implicits.toTraverseOps
```
kberec, line 19: Not needed. Can be removed in future branches if that's easier

app/viewmodels/models/TextAreaViewModel.scala
```diff
@@ -16,6 +16,12 @@
 
 package viewmodels.models
 
+import viewmodels.DisplayMessage.Message
+
 case class TextAreaViewModel(
-  rows: Int = 5
-)
+  rows: Int = 5,
+  hint: Option[Message] = None
```
kberec, line 23: It shouldn't break anything, but double check please, as it is used in multiple places. Check a11y too.

test/controllers/nonsipp/bondsdisposal/BondsDisposalListControllerSpec.scala
```diff
@@ -23,7 +23,13 @@ import eu.timepit.refined.refineMV
 import forms.RadioListFormProvider
 import models.{NormalMode, SchemeHoldBond}
 import pages.nonsipp.shares.SharesCompleted
```
kberec, line 25: Unused.  Can be fixed in future branches too.

kberec (approved):

PR #595, opened 2024-03-12
PSR-979 Quoted shares transformations

kberec:


app/transformations/SharesTransformer.scala
```diff
@@ -438,11 +441,16 @@ class SharesTransformer @Inject() extends Transformer {
   ): Try[UserAnswers] = {
 
     val userAnswersOfShares = userAnswers.set(DidSchemeHoldAnySharesPage(srn), shares.optShareTransactions.isDefined)
+    val userAnswersWithQuotedShares = if (shares.optTotalValueQuotedShares.isDefined) {
```
kberec, line 444: Any suggestions how to refactor this to look nicer?

cyeniceri: I rebased on main, tried to refactor here a little bit. I think all looks good. If you are happy too, we can merge when QA is done.

cyeniceri: We can also squash all these commits before merge in.

cyeniceri (dismissed):

cyeniceri (approved):

PR #599, opened 2024-03-13
PSR-816 Fixed employers contributions journey

westwater:


test/resources/logback.xml
```diff
@@ -1,10 +1,12 @@
 <configuration>
-    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
```
kberec, line 3: Is this meant to be changed?

app/controllers/PSRController.scala
```diff
@@ -107,8 +108,12 @@ abstract class PSRController extends FrontendBaseController with I18nSupport {
         .map(t => t._2.toString -> t._1)
         .toMap
 
-    def zipWithRefinedIndex[I: Validate[Int, *]]: List[(Refined[Int, I], A)] =
+    // todo: Seems unsafe, indexes that can't be refined get suppressed (they do not get added to the list)
```
kberec, line 111: Is this TODO for later of for this ticket?

test/controllers/nonsipp/employercontributions/EmployerContributionsMemberListControllerSpec.scala
```diff
@@ -56,25 +71,25 @@ class EmployerContributionsMemberListControllerSpec extends ControllerBaseSpec {
...
-    })
-
-    act.like(redirectNextPage(onSubmit, userAnswers, "value" -> "true"))
-    act.like(redirectNextPage(onSubmit, userAnswers, "value" -> "false"))
-
-    act.like(journeyRecoveryPage(onPageLoad).updateName("onPageLoad" + _))
-
-    act.like(invalidForm(onSubmit, userAnswers))
-    act.like(journeyRecoveryPage(onSubmit).updateName("onSubmit" + _))
+//    act.like(renderPrePopView(onPageLoad, EmployerContributionsMemberListPage(srn), true, userAnswers) {
```
kberec, line 78: FYI

app/controllers/PSRController.scala
westwater: changed to deprecated

westwater: changing to deprecated kills the compiler and forces code change, left as todo for now

app/controllers/nonsipp/employercontributions/package.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import scala.concurrent.Future
+
+package object employercontributions {
+
+  implicit class CallOps(call: Call) {
+
+    val isCyaPage: Boolean = {
+      val pattern =
+        s"^\\/pension-scheme-return\\/[^\\/]+\\/(change|check-answers)-employer-contributions\\/(?!.*\\/\\d+\\/\\d+$$).*"
```
kberec, line 35: Is there any risk that this is not going to work in the cloud environments? In case there are any issues with this we may need to choose some different way to identify the CYA page.

kberec (dismissed):

app/controllers/nonsipp/employercontributions/package.scala
westwater: I'm matching using the `url` value of the `Call` object which removes the host so should be fine

kberec (approved):

PR #610, opened 2024-03-18
PSR-866 Income from asset + WhyDoesSchemeHoldAssets cleanup

B-Savage:


app/controllers/nonsipp/otherassetsheld/IncomeFromAssetController.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+}
+
+object IncomeFromAssetController {
+  def form(formProvider: MoneyFormProvider): Form[Money] = formProvider(
+    MoneyFormErrors(
+      "otherAssets.incomeFromAsset.error.required",
+      "otherAssets.incomeFromAsset.error.invalid",
+      (Constants.maxMoneyValue, "otherAssets.incomeFromAsset.error.tooLarge"),
+      (Constants.zeroMoneyValue, "otherAssets.incomeFromAsset.error.tooSmall")
```
ccatosdevelopment, line 100: I'm wondering if it would be better to reuse `"otherAssets.incomeFromAsset.error.invalid"` here instead of `"otherAssets.incomeFromAsset.error.tooSmall"`, as the latter doesn't match a key in `messages.en` - what do you think? An alternative would be to create the key `"otherAssets.incomeFromAsset.error.tooSmall"` with the same error message as `"otherAssets.incomeFromAsset.error.invalid"`.

I appreciate this is sort of irrelevant, as any negative input will trigger the invalid input error message anyways, so this definitely shouldn't block this ticket from being merged in, but making a reference to a key that doesn't exist in `messages.en` stuck out to me, so I thought I'd raise this question.

ccatosdevelopment (approved):

app/controllers/nonsipp/otherassetsheld/IncomeFromAssetController.scala
B-Savage: I agree with reusing `otherAssets.incomeFromAsset.error.invalid`, and will correct this in the current branch I am working on.

PR #627, opened 2024-03-26
PSR-868 | OtherAssetsCYA

ccatosdevelopment:


app/controllers/nonsipp/common/IdentityTypeController.scala
```diff
@@ -62,6 +64,18 @@ class IdentityTypeController @Inject()(
     subject match {
       case IdentitySubject.Unknown => Redirect(controllers.routes.UnauthorisedController.onPageLoad())
       case _ =>
+        // If this page is reached in CheckMode, in Other Assets Journey, and there is no PointOfEntry set
+        if (mode == CheckMode && subject == IdentitySubject.OtherAssetSeller && request.userAnswers
```
kberec, line 68: One for later @ccatosdevelopment, shouldn't block the merge:  this is a very specific condition and IdentityTypeController is a very generic page. Is it possible to put this logic somewhere else?

narendrapaduchuri (approved):

narendrapaduchuri (approved):

app/controllers/nonsipp/common/IdentityTypeController.scala
ccatosdevelopment: Good point, and good question - if we can set the PointOfEntry when the user clicks the Change link on the CYA page, that would mean we can remove it from here, but off the top of my head, I'm not sure if/how it could be implemented this way. What do you think?

PR #635, opened 2024-04-02
PSR-956 any part asset still held page

narendrapaduchuri:


app/pages/nonsipp/otherassetsdisposal/AnyPartAssetStillHeldPage.scala
```diff
@@ -0,0 +1,32 @@
...
+import config.Refined.{Max50, Max5000}
+import pages.QuestionPage
+
+case class AnyPartAssetStillHeldPage(srn: Srn, assetIndex: Max5000, disposalIndex: Max50)
+    extends QuestionPage[Boolean] {
+
+  override def path: JsPath =
+    Paths.assetsDisposed \ toString \ assetIndex.arrayIndex.toString \ disposalIndex.arrayIndex.toString
+
+  override def toString: String = "portionStillHeld"
```
kberec, line 31: For awareness @narendrapaduchuri @cyeniceri, the schema name for this field is `assets.otherAssets.otherAssetTransactions[n].assetsDisposed[n].fullyDisposedOf` (and it means the opposite than the UI collects here). It is still fine IMO to have it as @narendrapaduchuri created `AnyPartAssetStillHeldPage, portionStillHeld`, but when we do the BE transformation, the value should be negated.

app/pages/nonsipp/otherassetsdisposal/AnyPartAssetStillHeldPage.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package pages.nonsipp.otherassetsdisposal
+
+import utils.RefinedUtils._
+import models.SchemeId.Srn
+import play.api.libs.json.JsPath
+import config.Refined.{Max50, Max5000}
+import pages.QuestionPage
+
+case class AnyPartAssetStillHeldPage(srn: Srn, assetIndex: Max5000, disposalIndex: Max50)
```
kberec, line 25: @narendrapaduchuri I'm not sure if there is an AC for this, but do we have to do the following?
If any asset disposal of an asset has a value of 'false' for AnyPartAssetStillHeldPage, the asset should not be visible on the 'Start reporting an assey disposal' page .

kberec (approved):

kberec (approved):

app/pages/nonsipp/otherassetsdisposal/AnyPartAssetStillHeldPage.scala
kberec: This is wrong now, @narendrapaduchuri you renamed `portionStillHeld` to `fullyDisposedOf`. Page name is still `AnyPartAssetStillHeldPage`. Let's have a chat.

kberec (changes requested):
Need to sort out page logic.

app/pages/nonsipp/otherassetsdisposal/AnyPartAssetStillHeldPage.scala
kberec: As discussed with @narendrapaduchuri , he is going to revert naming `fullyDisposedOf` in the next PR.

kberec (approved):

PR #636, opened 2024-04-03
PSR-1010 PSR transformation of other assets data - GET and POST

cyeniceri:


app/models/backend/responses/PensionSchemeReturnResponse.scala
```diff
@@ -1,61 +0,0 @@
...
- * limitations under the License.
- */
-
-package models.backend.responses
-
-import utils.WithName
-import models.backend.responses.DataEntryRule.{Fixed, Updated}
-import play.api.libs.json._
-
-case class PensionSchemeReturn(
```
kberec, line 23: It's weird that this was added in very early stages, when we had no information about the BE yet. FYI @narendrapaduchuri

kberec (approved):

PR #669, opened 2024-04-30
Psr-751 UI Error scenarios

kberec:


app/controllers/JourneyRecoveryController.scala
```diff
@@ -18,37 +18,52 @@ package controllers
...
 
-  def onPageLoad(continueUrl: Option[RedirectUrl] = None): Action[AnyContent] = identify { implicit request =>
-    val safeUrl: Option[String] = continueUrl.flatMap { unsafeUrl =>
-      unsafeUrl.getEither(OnlyRelative) match {
-        case Right(safeUrl) =>
-          Some(safeUrl.url)
-        case Left(message) =>
-          logger.info(message)
-          None
+  def onPageLoad(continueUrl: Option[RedirectUrl] = None, answersSavedDisplayVersion: Int = 0): Action[AnyContent] =
```
westwater, line 43: I'm not super clear on what `answersSavedDisplayVersion` does

kberec: Which of these two to render:
```
journeyRecovery.guidance2.version1 = We have saved your answers.
journeyRecovery.guidance2.version2 = We have saved your answers in this section until you are signed out.
```
I tried to do it with enum, but it got complicated so I went with Int up for now.

app/services/PsrSubmissionService.scala
```diff
@@ -55,13 +56,19 @@ class PsrSubmissionService @Inject()(
 
   def submitPsrDetailsWithUA(
     srn: Srn,
-    userAnswers: UserAnswers
+    userAnswers: UserAnswers,
+    optFallbackCall: Option[Call] = None
```
cyeniceri, line 60: Why are we using this fallbackCall as an option ? Couldn't see anywhere sent None.

cyeniceri (commented):
In my opinion, we should do this error handling on Controller level instead in connector.
I mean, send the response from the corresponding controller, get response code after executing API call, and decide what to do according to the response code in same controller. 

To unblock other tickets, we can create a different ticket and refactor here which I suggest.

app/services/PsrSubmissionService.scala
kberec: As this was introduced gradually across all the 28 commits :) 
We can refactor later if you think that is useful.

kberec:
> In my opinion, we should do this error handling on Controller level instead in connector. I mean, send the response from the corresponding controller, get response code after executing API call, and decide what to do according to the response code in same controller.
> 
> To unblock other tickets, we can create a different ticket and refactor here which I suggest.

Yes, we could do that as a refactoring. In that case each controller code that calls this should be re-written. We would also lose the central place to handle BE errors (ErrorsHandler.onServerError).

cyeniceri:
> > In my opinion, we should do this error handling on Controller level instead in connector. I mean, send the response from the corresponding controller, get response code after executing API call, and decide what to do according to the response code in same controller.
> > To unblock other tickets, we can create a different ticket and refactor here which I suggest.
> 
> Yes, we could do that as a refactoring. In that case each controller code that calls this should be re-written. We would also lose the central place to handle BE errors (ErrorsHandler.onServerError).

ErrorsHandler will stay same, we just throw the relevant exception with relevant parameters in each controller. Rest will be done by ErrorsHandler centrally

cyeniceri (approved):

PR #674, opened 2024-05-07
PSR-919 data displayed as user entered

narendrapaduchuri:


app/controllers/nonsipp/bondsdisposal/ReportBondsDisposalListController.scala
```diff
@@ -188,12 +188,14 @@ class ReportBondsDisposalListController @Inject()(
...
       .map {
         case indexes @ (index, _) =>
-          request.userAnswers
+          index -> request.userAnswers
             .get(BondsCompleted(srn, index))
             .getOrRecoverJourney
             .map(bondsDisposal => (indexes, bondsDisposal))
       }
       .toList
+      .sortBy { case (index, _) => index.value }
```
kberec, line 197: This sorts the primary index but it doesn't sort the secondary index (disposal index).
There is however an additional change coming in in the future to convert the reported disposal list page to a nested version (I think PSR-1018), so it might be not worthwhile to sort by the secondary index too, unless it is a very quick fix.
@narendrapaduchuri please check if it is a very quick fix and only implement if it is. FYI @BabuRupoll7623

app/controllers/nonsipp/bonds/BondsListController.scala
```diff
@@ -131,13 +131,16 @@ class BondsListController @Inject()(
...
-    indexes.map { index =>
-      for {
-        nameOfBonds <- requiredPage(NameOfBondsPage(srn, index))
-        acquisitionType <- requiredPage(WhyDoesSchemeHoldBondsPage(srn, index))
-        costOfBonds <- requiredPage(CostOfBondsPage(srn, index))
-      } yield BondsData(index, nameOfBonds, acquisitionType, costOfBonds)
-    }.sequence
+    indexes
+      .sortBy(x => x.value)
+      .map { index =>
```
kberec, line 136: Is there an easy way to test the sorting in the unit tests @narendrapaduchuri ?

app/controllers/nonsipp/bondsdisposal/ReportBondsDisposalListController.scala
BabuRupoll7623: Thanks @kberec , i have raised this issue with @narendrapaduchuri , though of checking with Design team on it

narendrapaduchuri: Hi @kberec , it doesn't help any way this extra bit of code. I think  we can merge this as it is. If you are ok with it.

kberec (approved):

PR #677, opened 2024-05-08
PSR-953 PSR-954 Removing member payments upon member deletion & fixing index dependencies

B-Savage:


app/pages/nonsipp/membersurrenderedbenefits/SurrenderedBenefitsAmountPage.scala
```diff
@@ -17,16 +17,32 @@
...
 
+import scala.util.Try
+
 case class SurrenderedBenefitsAmountPage(srn: Srn, memberIndex: Max300) extends QuestionPage[Money] {
 
   override def path: JsPath = Paths.memberPensionSurrender \ toString \ memberIndex.arrayIndex.toString
 
   override def toString: String = "totalSurrendered"
 
+  override def cleanup(value: Option[Money], userAnswers: UserAnswers): Try[UserAnswers] =
```
kberec, line 36: Is there a chance that this cleanup breaks the removal through `RemoveSurrenderedBenefitsController`? (Same question applies for all member payment cleanup methods)

app/controllers/nonsipp/memberdetails/upload/FileUploadSuccessController.scala
```diff
@@ -88,7 +99,21 @@ class FileUploadSuccessController @Inject()(
       UserAnswers.remove(DoesMemberHaveNinoPages(srn)),
       UserAnswers.remove(MemberDetailsNinoPages(srn)),
       UserAnswers.remove(NoNinoPages(srn)),
-      UserAnswers.remove(MemberStatuses(srn))
+      UserAnswers.remove(MemberStatuses(srn)),
```
kberec, line 102: FYI @westwater this PR may gets merged before the soft delete changes. Can you see anything that wouldn't fit with your approach of soft delete?

app/pages/nonsipp/memberdetails/MemberDetailsPage.scala
```diff
@@ -36,16 +43,65 @@ case class MemberDetailsPage(srn: Srn, index: Max300) extends QuestionPage[NameD
...
     }
 
-  private def pages(srn: Srn): List[Removable[_]] =
+  private def pages(srn: Srn, userAnswers: UserAnswers): List[Removable[_]] =
     List(
       DoesMemberHaveNinoPage(srn, index),
       MemberStatus(srn, index)
     )
 
+  private def paymentsPages(srn: Srn, index: Max300, userAnswers: UserAnswers): List[Removable[_]] = {
```
kberec, line 56: What should happen if the last member is removed? Should member payments journey type pages get removed (e.g.  TransfersInJourneyStatus) so that the tasklist and the data reverts back to the state how it was before adding any members?

app/controllers/nonsipp/memberdetails/upload/FileUploadSuccessController.scala
westwater: Yeah It looks like this will conflict with soft delete - because of how the soft delete is working, I need to hard delete all member journeys when a member gets deleted which look like part of what this PR is trying to do

kberec: Thanks @westwater . Do you want @B-Savage to go-ahead with these changes as they are and you're going to adjust them to the soft delete, or do you want @B-Savage to change the approach?
AFAIK hard delete of members is still valid in some scenarios (e.g. compiled status, version1)

app/pages/nonsipp/memberdetails/MemberDetailsPage.scala
B-Savage: I've added a solution to this now - all member related pages that are remaining are removed upon the last member being deleted.

app/controllers/nonsipp/memberdetails/upload/FileUploadSuccessController.scala
westwater: Yeah lets get this in and i'll deal with it on my PR

kberec (approved):

PR #678, opened 2024-05-10
Psr 453 Additional fixes for CIP Compile event

kberec:


app/services/PsrSubmissionService.scala
```diff
@@ -97,29 +100,36 @@ class PsrSubmissionService @Inject()(
...
+                optOtherAssetsDisposal
+              ),
+              membersPayments = memberPaymentsTransformer.transformToEtmp(srn, request.userAnswers),
+              shares = buildShares(srn)(optDidSchemeHoldAnyShares, optSharesDisposal),
+              psrDeclaration = Option.when(isSubmitted)(declarationTransformer.transformToEtmp)
+            )
+          )
+          .flatMap {
+            case Left(message: String) =>
+              throw PostPsrException(message, optFallbackCall.fold(appConfig.urls.managePensionsSchemes.dashboard) {
```
cyeniceri, line 125: I think optFallbackCall shouldn't be optional. All the controller where used sends fallback url inside Some, We can send it directly, and don't need to fold it later here

app/connectors/PSRConnector.scala
```diff
@@ -45,22 +44,18 @@ class PSRConnector @Inject()(appConfig: FrontendAppConfig, http: HttpClient) {
...
       .map { response =>
         response.status match {
-          case NO_CONTENT => NoContent
+          case NO_CONTENT => Right(true)
           case _ =>
-            val url = optFallbackCall.fold(appConfig.urls.managePensionsSchemes.dashboard) { fallbackCall =>
-              fallbackCall.url
-            }
-            throw PostPsrException(s"{${response.status}, ${response.json}}", url)
+            Left(s"{${response.status}, ${response.json}}")
```
cyeniceri, line 58: Just an idea, would it be better if we return Future[Either[String, Unit]] ? Because that boolean doesn't make sense to me.

```suggestion
              def submitPsrDetails(
    psrSubmission: PsrSubmission
  )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[String, Unit]] =
    http
      .POST[PsrSubmission, HttpResponse](
        submitStandardUrl,
        psrSubmission
      )
      .map { response =>
        response.status match {
          case NO_CONTENT => Right(())
          case _ =>
            Left(s"{${response.status}, ${response.json}}")
        }
      }
```

```
            case Right(()) =>
              auditService.sendExtendedEvent(buildAuditEvent(taxYear, loggedInUserNameOrBlank(request)))
              Future.unit
```

kberec: Yes, that makes sense, I'm going to change that.

app/services/PsrSubmissionService.scala
kberec: I'll check how big refactoring it would be.

app/connectors/PSRConnector.scala
```diff
@@ -45,15 +45,15 @@ class PSRConnector @Inject()(appConfig: FrontendAppConfig, http: HttpClient) {
...
+  )(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[String, Unit]] =
     http
       .POST[PsrSubmission, HttpResponse](
         submitStandardUrl,
         psrSubmission
       )
       .map { response =>
         response.status match {
-          case NO_CONTENT => Right(true)
+          case NO_CONTENT => Right(Future.unit)
```
cyeniceri, line 56: why Future.unit ? 
this is `final val unit: Future[Unit] = fromTry(Success(()))`

kberec: I thought they are equivalent. But I can change it, no problems.

app/services/PsrSubmissionService.scala
kberec: It's a big refactoring, but done it anyway, as we are at the beginning of the sprint.

app/connectors/PSRConnector.scala
kberec: I changed this.

cyeniceri (approved):

PR #680, opened 2024-05-14
PSR-892 PSR-890 Remove 'You must tell us all' text when max limit reached

B-Savage:


app/controllers/nonsipp/landorproperty/LandOrPropertyListController.scala
```diff
@@ -118,6 +118,7 @@ object LandOrPropertyListController {
 
     val title = if (addresses.size == 1) "landOrPropertyList.title" else "landOrPropertyList.title.plural"
     val heading = if (addresses.size == 1) "landOrPropertyList.heading" else "landOrPropertyList.heading.plural"
+    val paragraph = if (addresses.size < 5000) Some(ParagraphMessage("landOrPropertyList.paragraph")) else None
```
kberec, line 121: I think we have a constant for this. maxLandOrProperties

app/controllers/nonsipp/loansmadeoroutstanding/LoansListController.scala
```diff
@@ -156,6 +156,7 @@ object LoansListController {
 
     val title = if (recipients.length == 1) "loansList.title" else "loansList.title.plural"
     val heading = if (recipients.length == 1) "loansList.heading" else "loansList.heading.plural"
+    val description = if (recipients.length < 5000) Some(ParagraphMessage("loansList.description")) else None
```
kberec, line 159: I think we have a constant: `maxLoans`

app/controllers/nonsipp/landorproperty/LandOrPropertyListController.scala
```diff
@@ -118,6 +118,7 @@ object LandOrPropertyListController {
 
     val title = if (addresses.size == 1) "landOrPropertyList.title" else "landOrPropertyList.title.plural"
     val heading = if (addresses.size == 1) "landOrPropertyList.heading" else "landOrPropertyList.heading.plural"
```
kberec, line 120: Just to double-check, is there any other journey (e.g. shares, other assets, etc) where the same UI logic should be applied and it is not already there?

kberec (approved):

PR #682, opened 2024-05-14
PSR-405 | Page Title "Error:" Prefix

ccatosdevelopment:


app/controllers/nonsipp/shares/HowManySharesController.scala
```diff
@@ -76,7 +77,8 @@ class HowManySharesController @Inject()(
               Future.successful(
                 BadRequest(
                   view(
-                    HowManySharesController.viewModel(srn, index, nameOfSharesCompany, mode, formWithErrors)
+                    formWithErrors,
+                    HowManySharesController.viewModel(srn, index, nameOfSharesCompany, mode, form)
```
kberec, line 81: Does it matter that `form` is different from `formWithErrors` here (and probs in some other places too)?

test/views/IntViewSpec.scala
```diff
@@ -50,15 +50,15 @@ class IntViewSpec extends ViewSpec with Mappings {
...
-      act.like(renderInputWithLabel(tripleIntViewModel)("value.1", view(_), _.page.fields.head.label))
-      act.like(renderInputWithLabel(tripleIntViewModel)("value.2", view(_), _.page.fields(1).label))
-      act.like(renderInputWithLabel(tripleIntViewModel)("value.3", view(_), _.page.fields(2).label))
-      act.like(renderErrors(invalidViewModel)(view(_), _ => "int.error.required"))
-      act.like(renderForm(viewModelGen)(view(_), _.onSubmit))
-      act.like(renderButtonText(viewModelGen)(view(_), _.buttonText))
+      act.like(renderTitle(viewModelGen)(view(intForm, _), _.title.key))
+      act.like(renderHeading(viewModelGen)(view(intForm, _), _.heading))
+      act.like(renderInputWithLabel(singleIntViewModel)("value", view(intForm, _), _.heading))
+//      act.like(renderInputWithLabel(tripleIntViewModel)("value.1", view(tripleIntForm, _), _.page.fields.head.label))
```
kberec, line 56: Is this commented out code the one that's going to be sorted out as a separate PR?

app/controllers/nonsipp/landorproperty/LandOrPropertyTotalIncomeController.scala
```diff
@@ -58,7 +58,7 @@ class LandOrPropertyTotalIncomeController @Inject()(
     implicit request =>
       request.userAnswers.get(LandOrPropertyChosenAddressPage(srn, index)).getOrRecoverJourney { address =>
         val preparedForm = request.userAnswers.fillForm(LandOrPropertyTotalIncomePage(srn, index), form)
-        Ok(view(viewModel(srn, index, preparedForm, address.addressLine1, mode)))
+        Ok(view(preparedForm, viewModel(srn, index, form, address.addressLine1, mode)))
```
kberec, line 61: Same question as previously for a slightly different scenario: should `form` parameter value of viewModel be `preparedForm` here and in similar code changes?

test/views/MoneyViewSpec.scala
```diff
@@ -50,17 +48,19 @@ class MoneyViewSpec extends ViewSpec with Mappings {
...
-      act.like(renderInputWithLabel(tripleMoneyQuestion)("value.1", view(_), _.page.fields.head.label))
-      act.like(renderInputWithLabel(tripleMoneyQuestion)("value.2", view(_), _.page.fields(1).label))
-      act.like(renderInputWithLabel(tripleMoneyQuestion)("value.3", view(_), _.page.fields(2).label))
-      act.like(renderErrors(invalidSingleMoneyQuestion)(view(_), _ => "money.error.required"))
-      act.like(renderForm(viewModelGen)(view(_), _.onSubmit))
-      act.like(renderButtonText(viewModelGen)(view(_), _.buttonText))
+      act.like(renderTitle(viewModelGen)(view(moneyForm, _), _.title.key))
+      act.like(renderHeading(viewModelGen)(view(moneyForm, _), _.heading))
+      act.like(renderInputWithLabel(singleMoneyQuestion)("value", view(moneyForm, _), _.heading))
+//      act.like(renderInputWithLabel(tripleMoneyQuestion)("value.1", view(_), _.page.fields.head.label))
```
kberec, line 58: Same question for commented code as previously (checking commit by commit, may be resolved later)

app/controllers/nonsipp/bondsdisposal/TotalConsiderationSaleBondsController.scala
```diff
@@ -44,7 +45,7 @@ class TotalConsiderationSaleBondsController @Inject()(
   saveService: SaveService,
   @Named("non-sipp") navigator: Navigator,
   identifyAndRequireData: IdentifyAndRequireData,
-  formProvider: MoneyFormErrorProvider,
+  formProvider: MoneyFormProvider,
```
kberec, line 48: Should we do this change in .g8 files too and remove MoneyFormErrorProvider?

test/views/IntViewSpec.scala
ccatosdevelopment: Yes, these tests will be replaced by `MultipleQuestionViewSpec`, on a separate branch (e.g. `PSR-405-unit-tests`)

app/controllers/nonsipp/landorproperty/LandOrPropertyTotalIncomeController.scala
ccatosdevelopment: No, for the same reason as above:

"No, this is intentional - the `form` parameter in the `viewModel` method call is used only to satisfy further method calls and constructors, as it needs the form type to do so (e.g. in `FormPageViewModel` and `SingleQuestion`). The `form` parameter in the `view` method call is what's used to populate the page with any pre-populated data."

test/views/MoneyViewSpec.scala
ccatosdevelopment: As above - yes, these tests will be replaced by `MultipleQuestionViewSpec`, on a separate branch (e.g. `PSR-405-unit-tests`)

app/controllers/nonsipp/bondsdisposal/TotalConsiderationSaleBondsController.scala
ccatosdevelopment: I think that's a good idea, unless there are concerns it might break things, or incur significant re-testing efforts to include that in this ticket. Or would this change be done as part of a separate branch/ticket?

kberec: Yes, in the follow up branch with unit tests please.

kberec (approved):

kberec (approved):

PR #683, opened 2024-05-14
PSR-636 Soft delete members

westwater:


app/utils/nonsipp/SoftDelete.scala
```diff
@@ -0,0 +1,235 @@
...
+      get(MemberDetailsNinoPage).map(_.value).some
+    ).mapN(
+      (nameDob, noNino, nino) => MemberPersonalDetails(nameDob.firstName, nameDob.lastName, nino, noNino, nameDob.dob)
+    )
+
+    val memberDetailsPages: List[Removable[_]] = List(
+      MemberDetailsPage(srn, index),
+      DoesMemberHaveNinoPage(srn, index),
+      NoNINOPage(srn, index),
+      MemberDetailsNinoPage(srn, index)
```
kberec, line 69: There may be pages that are missing from here, when I look at the data after the soft delete, some are still there. E.g. memberStatus, reportAnotherTransferIn, totalMemberContribution, memberLumpSumReceived, pensionAmountReceived

app/controllers/nonsipp/memberdetails/RemoveMemberDetailsController.scala
```diff
@@ -72,11 +75,16 @@ class RemoveMemberDetailsController @Inject()(
           removeMemberDetails => {
             if (removeMemberDetails) {
               for {
-                updatedAnswers <- Future.fromTry(request.userAnswers.remove(MemberDetailsPage(srn, index)))
+                updatedAnswers <- softDeleteMember(srn, index).mapK[Future]
```
kberec, line 78: I some cases, e.g, when the user start return in compile mode and still hasn't done the declaration, members should be hard deleted. Is there a logic to tell when to hard delete and when to soft delete?

kberec (commented):
Appreciate it isn't done yet, just some initial comments

app/controllers/nonsipp/memberdetails/RemoveMemberDetailsController.scala
westwater: No logic like that, I can change the softDeletemember to just deleteMember and pass in a softDelete flag - is there a flag that is set when the the return is in compile or submitted mode?

app/utils/nonsipp/SoftDelete.scala
westwater: will add 👍

app/controllers/nonsipp/memberdetails/RemoveMemberDetailsController.scala
kberec: I think there is a need for that. Not sure about the exact conditions though. Maybe when the return main version is 001. There were some conversations about it, maybe Stuart can help.

westwater: would you be ok if we open up a story for this and tackle it when I get back? should just be able to layer that functionality in quite nicely

kberec: Yes, I'll create a story

app/utils/nonsipp/SoftDelete.scala
westwater: added the extra pages to remove

kberec: Looking good, all are now removed as far as I see, except "employerContributionsCompleted". Should that be removed too?

app/transformations/MemberPaymentsTransformer.scala
```diff
@@ -88,6 +86,21 @@ class MemberPaymentsTransformer @Inject()(
         )
     }
 
+    val softDeletedMembers: List[MemberDetails] = userAnswers.get(SoftDeletedMembers(srn)).toList.flatten.map {
```
kberec, line 89: There is some problem here. If a new member is filling the index gap of a soft deleted member, list of softDeletedMembers won't contain the new member. More details on slack.

kberec: @BabuRupoll7623 cannot replicate this, so it might be a false alarm.

kberec (approved):

PR #693, opened 2024-05-23
PSR-743 Don't POST heldProperty.otherDescription when empty

B-Savage:


app/transformations/LandOrPropertyTransactionsTransformer.scala
```diff
@@ -548,7 +548,7 @@ class LandOrPropertyTransactionsTransformer @Inject() extends Transformer {
                       identityType = receivedLandType,
                       idNumber = None,
                       reasonNoIdNumber = None,
-                      otherDescription = Some(other.description)
+                      otherDescription = if (other.description.nonEmpty) Some(other.description) else None
```
cyeniceri, line 551: Could be rewritten as 
`
Option.when(other.description.nonEmpty)(other.description)
`

cyeniceri (approved):

kberec (approved):

kberec (changes requested):
Looks @harishmuddegowda found some issue when testing, after the data is posted and subsequently fetched, please refer to the slack conversation.

app/transformations/LandOrPropertyTransformer.scala
```diff
@@ -211,7 +211,7 @@ class LandOrPropertyTransformer @Inject() extends Transformer {
               prop => {
                 OtherRecipientDetailsPage(srn, index, LandOrPropertySeller) -> RecipientDetails(
                   heldPropertyTransaction.optPropertyAcquiredFromName.get,
-                  prop.otherDescription.get
+                  prop.otherDescription.getOrElse(None)
```
cyeniceri, line 214: would it be better if we use :
prop.otherDescription instead prop.otherDescription.getOrElse(None)

cyeniceri: now it's changed :)

cyeniceri (approved):

kberec (approved):

PR #696, opened 2024-05-24
PSR-1174 Email content

kberec:


app/controllers/nonsipp/declaration/PsaDeclarationController.scala
```diff
@@ -67,12 +68,14 @@ class PsaDeclarationController @Inject()(
       schemeDateService.schemeDate(srn) match {
         case None => Future.successful(Redirect(controllers.routes.JourneyRecoveryController.onPageLoad()))
         case Some(dates) =>
+          val now = schemeDateService.now()
           def emailFuture: Future[EmailStatus] =
```
cyeniceri, line 72: Do we really need this "extra" emailFuture here ? It seems it's just calling sendEmail only.

app/controllers/nonsipp/declaration/PsaDeclarationController.scala
```diff
@@ -86,25 +89,29 @@ class PsaDeclarationController @Inject()(
...
             Redirect(navigator.nextPage(PsaDeclarationPage(srn), NormalMode, request.userAnswers))
               .addingToSession((RETURN_PERIODS, schemeDateService.returnPeriodsAsJsonString(srn)))
-              .addingToSession((SUBMISSION_DATE, schemeDateService.submissionDateAsString(schemeDateService.now())))
+              .addingToSession((SUBMISSION_DATE, schemeDateService.submissionDateAsString(now)))
           }
       }
     }
 
-  private def sendEmail(name: String, email: String, taxYear: DateRange, schemeName: String)(
+  private def sendEmail(
```
cyeniceri, line 97: same comment for the emailFuture as above

app/controllers/nonsipp/declaration/PsaDeclarationController.scala
kberec: I deliberately left it as it is in ER.

kberec: We can refactor it later when everything works if you think it is necessary.

cyeniceri: so shall we remove it if it's not required then ?

cyeniceri (approved):

app/controllers/nonsipp/declaration/PsaDeclarationController.scala
cyeniceri: I think they are unnecessary :) okay, we can remove them later.

PR #718, opened 2024-06-07
PSR-1185 Added SRN to sign out and keep alive calls

kberec:


conf/application.conf
```diff
@@ -131,7 +131,7 @@ timeout-dialog {
 }
 
 session {
-  timeoutSeconds = 900
+  timeoutSeconds = 901
```
kberec, line 134: deliberate :)

B-Savage (approved):

PR #719, opened 2024-06-07
PSR-1007 Show disposal task list items conditionally

B-Savage:


app/utils/nonsipp/TaskListUtils.scala
```diff
@@ -387,23 +391,30 @@ object TaskListUtils {
...
+        messageKey("nonsipp.tasklist.sharesdisposal", "title", sharesDisposalsStatus),
+        sharesDisposalsLinkUrl
+      ),
+      sharesDisposalsStatus
+    )
+
+    val viewModelList = NonEmptyList
+      .fromList(userAnswers.get(SharesCompleted.all(srn)) match {
+        case Some(_) => List(sharesItem, sharesDisposalItem)
+        case None => List(sharesItem)
```
kberec, line 413: Is this supposed to be only when DidSchemeHoldAnySharesPage is not  Some(true) ? As when `SharesCompleted` is Some(_) , there still can be some share disposals? Happy to discuss....

kberec (approved):

PR #740, opened 2024-06-26
PSR-1226 transfers-in CYA view only page

narendrapaduchuri:


app/controllers/nonsipp/receivetransfer/TransfersInCYAController.scala
```diff
@@ -55,51 +58,82 @@ class TransfersInCYAController @Inject()(
...
-                )
-            )
-            .map {
-              case Nil => Redirect(controllers.routes.JourneyRecoveryController.onPageLoad())
-              case journeys => Ok(view(viewModel(srn, memberDetails.fullName, index, journeys, mode)))
-            }
-            .merge
-        }
-      ).merge
+      onPageLoadCommon(srn: Srn, index: Max300, mode: Mode)(implicitly)
```
kberec, line 61: Just spotted this, it was a mistake in the original example I think. Should this be: 
```
onPageLoadCommon(srn, index, mode)(implicitly)
```
? (same in onPageLoadViewOnly)

kberec: If you agree, could you please change it in basic / financial details too @narendrapaduchuri ?

kberec (dismissed):

app/controllers/nonsipp/receivetransfer/TransfersInCYAController.scala
narendrapaduchuri: Yes, @kberec I agree with you. I changed across above mentioned and  member contributions as well. Thanks

ccatosdevelopment (approved):

app/controllers/nonsipp/receivetransfer/TransfersInCYAController.scala
kberec: Thanks. Just the commits needs squashed and it is good to get merged.

kberec (approved):

PR #746, opened 2024-06-27
PSR-841 Trim whitespace on UTR

B-Savage:


cyeniceri (approved):

app/models/Utr.scala
```diff
@@ -24,7 +24,7 @@ case class Utr(utr: String) extends TaxIdentifier with SimpleName {
 
   override def toString = utr
 
-  def value = utr
+  def value: String = utr.filterNot(_.isWhitespace)
```
kberec, line 27: I'd consider either to add a unit test or at least to change the Utr generator to include space(s). Maybe something like:
```
  implicit val utrGen: Gen[Utr] = for {
    firstSetOfNumbers <- Gen.listOfN(5, Gen.numChar).map(_.mkString)
    secondSetOfNumbers <- Gen.listOfN(5, Gen.numChar).map(_.mkString)
  } yield {
    Utr(s"$firstSetOfNumbers $secondSetOfNumbers")
  }

```

PR #751, opened 2024-07-02
PSR-1164 Set member status to CHANGED on amending any member payments

westwater:


app/controllers/PSRController.scala
```diff
@@ -105,6 +107,10 @@ abstract class PSRController extends FrontendBaseController with I18nSupport {
         }
     }
 
+  def memberPaymentsChanged(implicit request: DataRequest[_]): Boolean = request.pureUserAnswers.exists(
```
kberec, line 110: Can this replace the one in `SchemeMemberDetailsAnswersController`?

app/controllers/PSRController.scala
```diff
@@ -105,6 +107,10 @@ abstract class PSRController extends FrontendBaseController with I18nSupport {
...
+    !_.sameAs(request.userAnswers, membersPayments, Omitted.membersPayments: _*)
```
kberec, line 111: Maybe I'm missing something here, but this doesn't consider the index of the member, so if you have multiple members and any of the members has changed and e.g. you navigate to CYA page for a member that hasn't been changed and submit the CYA page then memberStatus get flipped to "Changed" (and by that version for that member increased too).

westwater: Thats true but we omit MemberStatus from the comparison, it reality it's the old comparison logic that Cem wrote but I had to omit fields like MemberStatus for that very reason

In hindsight we probably should have built a case class with all member payments fields pulled from UA (like we do when we compare members for memberPsrVersion)

app/transformations/MemberPaymentsTransformer.scala
```diff
@@ -115,43 +116,85 @@ class MemberPaymentsTransformer @Inject()(
...
         case (index, currentMemberDetail) =>
           val optInitialMemberDetail = initialMemberDetails.get(index)
+          val refinedCurrentMember =
+            currentMemberDetail
+              .copy(
+                memberLumpSumReceived =
+                  if (currentMemberDetail.memberLumpSumReceived.exists(_.empty)) None
+                  else currentMemberDetail.memberLumpSumReceived,
+                pensionAmountReceived =
+                  if (currentMemberDetail.pensionAmountReceived.contains(0.0)) None
```
kberec, line 172: Above logic checks `.exists(_.empty)`. This one (and the one below) checks `.contains(0.0)`. Should these be the same?

app/controllers/PSRController.scala
kberec: I probably couldn't describe the problem properly, so let's have a chat about it.

app/transformations/MemberPaymentsTransformer.scala
westwater: .empty is just a function I added to the MemberLumpSumReceived case class, it does the same thing (checks both double values in the case class are 0.0)

app/controllers/PSRController.scala
westwater: Fixed

kberec: `PSRController.memberPaymentsChanged` still being called from multiple places in member payments and some member-index data set based on it. E.g.  `MemberPensionPaymentsCYAController.onSubmit`, `EmployerContributionsCYAController.onSubmit`, and 5 other places. Would we have a similar bug if e.g. member payment is CYA page is reached on the UI and submitted without a change?

westwater: Fixed now, thanks for that 👍

app/transformations/MemberPaymentsTransformer.scala
```diff
@@ -177,14 +179,15 @@ class MemberPaymentsTransformer @Inject()(
               )
           val same = optInitialMemberDetail.contains(refinedCurrentMember)
           if (!same) {
+//            println(s"======== i $optInitialMemberDetail")
```
kberec, line 182: Some left over println, can be removed later on in a separate PR if that speeds things up

kberec (commented):
Looks good, just some conflicts to sort out.

app/transformations/MemberPaymentsTransformer.scala
westwater: removed

kberec (dismissed):

kberec (approved):

PR #754, opened 2024-07-03
PSR-1220 unallocated contribution view only page

narendrapaduchuri:


test/controllers/nonsipp/memberpayments/UnallocatedContributionCYAControllerSpec.scala
```diff
@@ -88,4 +116,64 @@ class UnallocatedContributionCYAControllerSpec extends ControllerBaseSpec {
...
+              ViewOnlyMode,
+              viewOnlyUpdated = false,
+              optYear = Some(yearString),
+              optCurrentVersion = Some(submissionNumberTwo),
+              optPreviousVersion = Some(submissionNumberOne),
+              compilationOrSubmissionDate = Some(submissionDateTwo)
+            )
+          )
+      }.before(mockTaxYear(dateRange))
+        .withName("OnPageLoadViewOnly renders ok with no changed flag")
```
ccatosdevelopment, line 149: I think it would be good to also have a test for the page rendering correctly with the Updated flag, what do you think?

ccatosdevelopment (approved):

PR #774, opened 2024-07-16
PSR-1257 Select a member to report pension payments - T1 - (page 23b)…

cyeniceri:
… - View only mode


test/controllers/nonsipp/memberpensionpayments/MemberPensionPaymentsListControllerSpec.scala
```diff
@@ -102,4 +128,85 @@ class MemberPensionPaymentsListControllerSpec extends ControllerBaseSpec {
     act.like(invalidForm(onSubmit, userAnswers))
     act.like(journeyRecoveryPage(onSubmit).updateName("onSubmit" + _))
   }
+
+  "MemberPensionPaymentsListController in view only mode" - {
+
+    val currentUserAnswers = userAnswers
+      .unsafeSet(FbVersionPage(srn), "002")
+      .unsafeSet(CompilationOrSubmissionDatePage(srn), submissionDateTwo)
```
kberec, line 136: Is it ok not to have member details in current/previous user answers?

kberec (approved):

test/controllers/nonsipp/memberpensionpayments/MemberPensionPaymentsListControllerSpec.scala
cyeniceri: I'm already reusing userAnswers which has memberDetails set in class scope. I think it should be alright unless you think different.

kberec: Sorry, I missed out that detail :) It is all right

PR #780, opened 2024-07-19
PSR-1311 Part3 - fully remove HttpClient

kberec:


app/handlers/ErrorHandler.scala
```diff
@@ -82,4 +83,6 @@ class ErrorHandler @Inject()(
       case _ => super.onServerError(request, exception)
     }
   }
+
+  override protected implicit val ec: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
```
kberec, line 87: We'd still need to check this.

ccatosdevelopment (approved):

PR #782, opened 2024-07-22
PSR-1285 Display wound-up scheme on the overview page for the current…

cyeniceri:
… year

### DIFF
--- a/app/controllers/OverviewController.scala
+++ b/app/controllers/OverviewController.scala
@@ -52,7 +52,6 @@ class OverviewController @Inject()(
   val controllerComponents: MessagesControllerComponents,
   psrOverviewService: PsrOverviewService,
   psrVersionsService: PsrVersionsService,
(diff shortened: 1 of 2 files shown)


app/controllers/OverviewController.scala
```diff
@@ -265,12 +264,12 @@ class OverviewController @Inject()(
 
 object OverviewController {
 
-  def options(startingTaxYear: TaxYear): Enumerable[DateRange] = {
+  def options(startingTaxYear: TaxYear): List[(String, DateRange)] = {
 
-    val taxYears = List.iterate(startingTaxYear, 7)(_.previous)
+    val taxYears = startingTaxYear.next +: List.iterate(startingTaxYear, 7)(_.previous)
```
kberec, line 269: Do we need to change the methods in `SchemeDateServiceImpl` too?

kberec (approved):

PR #783, opened 2024-07-23
PSR-1319 Fixed QA bug

westwater:


app/transformations/MemberPaymentsTransformer.scala
```diff
@@ -156,35 +157,46 @@ class MemberPaymentsTransformer @Inject()(
...
+            case None =>
+              currentMemberDetail.copy(memberPSRVersion = None)
+            case Some(initialMemberDetail) =>
+              val normalisedInitialMember = normalise(initialMemberDetail)
+              val normalisedCurrentMember = normalise(currentMemberDetail)
+              val same = normalisedInitialMember == normalisedCurrentMember
+              if (!same) {
+                logger.info(s"member $index has changed, removing memberPSRVersion")
+                if(logger.isDebugEnabled) {
+                  logger.debug(Diff(normalisedInitialMember, normalisedCurrentMember).mkString(" - "))
```
westwater, line 193: prints human readable, field specific diffs - great for debugging issues in the future around recordVersion / memberPsrVersion

app/models/requests/psr/MemberPayments.scala
```diff
@@ -75,7 +75,11 @@ case class MemberLumpSumReceived(
   lumpSumAmount: Double,
   designatedPensionAmount: Double
 ) {
-  val zero: Boolean = lumpSumAmount == 0.0 && designatedPensionAmount == 0.0
+  val zero: Boolean = lumpSumAmount == 0d && designatedPensionAmount == 0d
```
kberec, line 78: Would it be useful to differentiate in naming between this `zero` and the `zero` below? I.e. rename this to `isZero` maybe?

app/transformations/MemberPaymentsTransformer.scala
```diff
@@ -156,35 +157,46 @@ class MemberPaymentsTransformer @Inject()(
         }
       }
 
+      val normalise = (memberDetails: MemberDetails) => {
```
kberec, line 160: If we change the task list not to distinguish between In progress and Completed, do we need to revert this change? (FYI @ccatosdevelopment )

scalastyle-config.xml
```diff
@@ -87,7 +87,7 @@
     </check>
     <check level="warning" class="org.scalastyle.scalariform.CyclomaticComplexityChecker" enabled="true">
         <parameters>
-            <parameter name="maximum"><![CDATA[30]]></parameter>
+            <parameter name="maximum"><![CDATA[35]]></parameter>
```
kberec, line 90: Even 30 is quite hight TBH....

app/transformations/MemberPaymentsTransformer.scala
westwater: Yeah we would, this is just to fix the QA bug

kberec (approved):

scalastyle-config.xml
westwater: Yeah it is but it's just the nature of the transformer - might be worth splitting out all the code into seperate methods and the core `toEtmp/fromEtmp` is just a simple for-comp

PR #784, opened 2024-07-23
PSR-1318 | New Task List Design

ccatosdevelopment:


app/viewmodels/models/TaskListViewModel.scala
```diff
@@ -79,4 +79,8 @@ object TaskListStatus {
   case object Completed extends TaskListStatus("tasklist.completed")
 
   case object Updated extends TaskListStatus("tasklist.updated")
+
+  case object Reported extends TaskListStatus("tasklist.reported")
+
+  case class Reported(amount: String) extends TaskListStatus(Message("tasklist.reportedWithAmount", amount))
```
kberec, line 85: Would it be better to keep `amount` as Int and just detect `0` when constructing the message and display `None` in that case?

app/controllers/nonsipp/ViewOnlyTaskListController.scala
```diff
@@ -611,12 +659,24 @@ object ViewOnlyTaskListController {
...
       Paths.shares \ "totalValueQuotedShares"
     )
 
+    val quotedSharesTaskListStatus = quotedSharesCompletedOrUpdated match {
+      case Completed =>
+        currentUA.get(TotalValueQuotedSharesPage(srn)) match {
+          case Some(amount) if !amount.isZero => Reported
+          // The condition above is necessary because there is no boolean field in ETMP where we can store the answer to
+          // the first question page, so a value of 0.00 stored in ETMP indicates that no Quoted Shares were reported.
+          case _ => Reported("None")
```
westwater, line 674: how is Reported("None") interpreted? Does it display none or is it consumed by some logic to output something else?

ccatosdevelopment: On the Task List page the status will display as "None reported", is that what you mean in terms of how things are interpreted? Or are you asking more about how our code handles the `Reported` status internally, rather than how it's rendered?

app/viewmodels/models/TaskListViewModel.scala
ccatosdevelopment: That was my initial approach, but I ran into hurdles with the implementation - I don't remember the exact circumstances, but it was something to do with having a combination of `case object`, `case class`, and/or `object` with an `apply` method that would carry out that `0` detection. There was always at least one error, and any fix I found to solve one error would result in a different error somewhere else.

Since it was beginning to get time-consuming, I opted for this implementation instead - I appreciate `amount: String` isn't great, but I think in this instance it's not too bad, because we don't do any further calculations to the value after it's generated, so we don't have to convert back and forth between `Int` and `String` multiple times, and deal with any of the headaches that might entail.

kberec: Not sure what those hurdles are, but I was thinking that if we do something like `case class Reported(amount: String) extends TaskListStatus(Message("tasklist.reportedWithAmount", if (amount == 0) "None" else amount.toString))
` that would simplify the calling code.
I.e.
`if (numReported == 0) Reported("None") else Reported(numReported.toString)` can be replaced with `Reported(numReported)`

ccatosdevelopment: Ah okay, I see what you mean - thanks for the suggestion, I've just implemented this and pushed the change!

app/controllers/nonsipp/ViewOnlyTaskListController.scala
westwater: Yeah it was more about how the code handles it - if it's comparing the string value and checking if it's the value "None", it might make more sense to have that value at the object level e.g. `Reported.None` or something like that - this is minor though but in my experience arbitrary string comparison could lead to issues with miss-spelling, case sensitivity, etc but this is just a nitpick

westwater (dismissed):

app/controllers/nonsipp/moneyborrowed/BorrowInstancesListController.scala
```diff
@@ -127,7 +127,6 @@ class BorrowInstancesListController @Inject()(
       }
```
ccatosdevelopment, line 127: Autoformatted file, unrelated to task list redesign.

test/controllers/nonsipp/moneyborrowed/BorrowInstancesListControllerSpec.scala
```diff
@@ -45,7 +45,8 @@ class BorrowInstancesListControllerSpec extends ControllerBaseSpec {
   private lazy val onSubmit =
```
ccatosdevelopment, line 45: Autoformatted file, unrelated to task list redesign.

test/controllers/nonsipp/loansmadeoroutstanding/LoansListControllerSpec.scala
```diff
@@ -221,23 +221,21 @@ class LoansListControllerSpec extends ControllerBaseSpec {
```
ccatosdevelopment, line 221: Autoformatted file, unrelated to task list redesign.

PR #795, opened 2024-07-30
PSR-1298 PoC for separate no view in same controller

kberec:


kberec:
Closed, as different approach has been taken

PR #803, opened 2024-08-05
PSR-1260 Added view only page for quoted shares

westwater:


app/controllers/nonsipp/totalvaluequotedshares/TotalValueQuotedSharesCYAController.scala
```diff
@@ -83,79 +134,139 @@ class TotalValueQuotedSharesCYAController @Inject()(
...
           schemeDetails,
-          CheckMode
+          mode
         )
       ),
       refresh = None,
-      buttonText = "site.saveAndContinue",
-      onSubmit = controllers.nonsipp.totalvaluequotedshares.routes.TotalValueQuotedSharesCYAController.onSubmit(srn)
+      buttonText =
+        mode.fold(normal = "site.saveAndContinue", check = "site.continue", viewOnly = "site.return.to.tasklist"),
```
ccatosdevelopment, line 176: According to the content doc, the button text should be "Save and continue" for both `NormalMode` and `CheckMode` - if that's right, maybe this could be amended in the next PR? In any case I'll approve the PR as it wouldn't make sense to hold up merging for something small like this.

ccatosdevelopment (approved):

app/controllers/nonsipp/totalvaluequotedshares/TotalValueQuotedSharesCYAController.scala
westwater: ah good spot! i'll amend in the next one 👍

PR #805, opened 2024-08-06
PSR-1186 | Implement changes

ccatosdevelopment:


app/controllers/nonsipp/ViewOnlyTaskListController.scala
```diff
@@ -632,7 +636,7 @@ object ViewOnlyTaskListController {
     val quotedSharesStatusAndLink: TaskListStatus = getCompletedOrUpdatedTaskListStatus(
       currentUA,
       previousUA,
-      pages.nonsipp.totalvaluequotedshares.Paths.quotedShares
+      pages.nonsipp.totalvaluequotedshares.Paths.quotedShares \ "totalValueQuotedShares"
```
ccatosdevelopment, line 639: This is necessary based on the way our implementation works around there being no field in ETMP for the Yes/No page answer (`quotedSharesManagedFundsHeld`). Because of this, we don't want that field included in this comparison, we only want to compare this monetary value.

app/controllers/nonsipp/memberpayments/UnallocatedContributionCYAController.scala
```diff
@@ -74,7 +73,7 @@ class UnallocatedContributionCYAController @Inject()(
             getCompletedOrUpdatedTaskListStatus(
               request.userAnswers,
               request.previousUserAnswers.get,
-              membersPayments \ "unallocatedContribAmount"
+              pages.nonsipp.memberpayments.Paths.membersPayments \ "unallocatedContribAmount"
```
ccatosdevelopment, line 76: While fixing a Path import mistake of my own, I noticed that the same mistake was made here, so I fixed this at the same time.

app/utils/nonsipp/TaskListStatusUtils.scala
```diff
@@ -1003,4 +1003,114 @@ object TaskListStatusUtils {
...
+      previousUA,
+      pages.nonsipp.shares.Paths.shareTransactions,
+      Some("disposedSharesTransaction")
+    ) == Completed) &&
+    (getCompletedOrUpdatedTaskListStatus(
+      currentUA,
+      previousUA,
+      pages.nonsipp.sharesdisposal.Paths.disposedSharesTransaction
+    ) == Completed) &&
+    (getCompletedOrUpdatedTaskListStatus(
```
kberec, line 1077: If there are no land or property items, there are no disposals (menu item hidden). Would this logic still work in that scenario?

conf/messages.en
```diff
@@ -67,9 +67,9 @@ identityType.company = A UK company
 identityType.partnership = A UK partnership
 identityType.other = Other
 
-tasklist.unableToStart = Cannot start yet
-tasklist.notStarted = Not started
-tasklist.inProgress = In progress
+tasklist.unableToStart = Cannot start yet
```
kberec, line 70: 👍

app/utils/nonsipp/TaskListStatusUtils.scala
kberec: FYI @BabuRupoll7623 could you please check if this is working as expected?

BabuRupoll7623: Hidden disposal link working fine for (LoP, Bonds , 
shares and other assets)

kberec (approved):

PR #821, opened 2024-08-15
PSR-1100 addressed basic details issues

westwater:


app/pages/nonsipp/BasicDetailsCompletedPage.scala
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package pages.nonsipp
+
+import pages.QuestionPage
+import models.SchemeId.Srn
+import play.api.libs.json.{__, JsPath}
+import viewmodels.models.SectionCompleted
+
+case class BasicDetailsCompletedPage(srn: Srn) extends QuestionPage[SectionCompleted] {
```
kberec, line 24: Do we need to set this value up in FE mongo cache on GET from ETMP?

app/controllers/nonsipp/CheckReturnDatesController.scala
```diff
@@ -77,11 +79,40 @@ class CheckReturnDatesController @Inject()(
...
           .fold(
             formWithErrors => Future.successful(BadRequest(view(formWithErrors, viewModel))),
-            value =>
-              for {
-                updatedAnswers <- Future.fromTry(request.userAnswers.set(CheckReturnDatesPage(srn), value))
-                _ <- saveService.save(updatedAnswers)
-              } yield Redirect(navigator.nextPage(CheckReturnDatesPage(srn), mode, updatedAnswers))
+            value => {
+
+              val updatedUserAnswers = request.userAnswers.get(CheckReturnDatesPage(srn)) match {
```
kberec, line 84: Quite a bit of logic added here... 
- would it be useful it this is moved to a separate method?
- do we have tests for this logic?

app/pages/nonsipp/BasicDetailsCompletedPage.scala
westwater: No, since we don't go back to amend basic details after a submission theres no need

kberec (dismissed):

kberec (approved):

kberec (approved):

app/controllers/nonsipp/CheckReturnDatesController.scala
westwater: I've shrunk the logic quite a bit here now and added unit tests

PR #829, opened 2024-08-20
PSR-1344 CIP - Explicit Audit Event: PensionSchemeReturnSubmitted

cyeniceri:


app/services/PsrSubmissionService.scala
```diff
@@ -86,26 +86,27 @@ class PsrSubmissionService @Inject()(
               ).mapN {
                 (minimalRequiredSubmission, checkReturnDates, taxYear) =>
                   {
+                    val submissionRequest: PsrSubmission = PsrSubmission(
```
kberec, line 89: `if (isAnyUpdated) ` check above currently prevents this block to be executed if the user logs in, doesn't change anything and just submits the declaration. We do have a ticket to hide the link if nothing is changed from the PREVIOUS submission, however would it be better to change the condition to: `if (isSubmitted || initialUA.data != currentUA.data) ` ?

cyeniceri: what would be the benefit of this additional condition `isSubmitted ||` in this check `if (isAnyUpdated)` ? You can't reach out Declaration page if you don't have any changes from BasicDetailsCYA. Also as you said above, we have another ticket to prevent user to access the Declaration page without any changes. 
If we don't have any other case to access the Declaration page, we can leave here as is IMO.

kberec: Currently, if you fill in everything, log out and then log in again, will this method be called.

cyeniceri: If you fill in everything, when you log out/in, current&initial ua will be same until you click to save and continue in declaration page, right? and if you declare current&initial will be different and this method will be called.

cyeniceri: enable/disable declaration page link shouldn't be (initialUA.data != currentUA.data) then. We need to check last submitted version and current one.

kberec: It's not (initialUA.data != currentUA.data) part , it is `isSubmitted`

kberec (approved):

kberec (approved):

PR #830, opened 2024-08-20
PSR-1364 Add change link to basic details

B-Savage:


app/controllers/nonsipp/BasicDetailsCheckYourAnswersController.scala
```diff
@@ -334,14 +334,16 @@ object BasicDetailsCheckYourAnswersController {
...
+              ).withChangeAction(controllers.nonsipp.routes.WhichTaxYearController.onPageLoad(srn, CheckMode).url)
+                .withOneHalfWidth()
             )
           case Right(accountingPeriods) =>
             List(
               CheckYourAnswersRowViewModel(
                 "basicDetailsCheckYourAnswersController.schemeDetails.taxYear",
                 whichTaxYearPage.get.show
-              ).withOneHalfWidth()
+              ).withChangeAction(controllers.nonsipp.routes.WhichTaxYearController.onPageLoad(srn, CheckMode).url)
```
kberec, line 345: Both should navigate to accounting periods, not to the tax year page

kberec (approved):

app/controllers/nonsipp/BasicDetailsCheckYourAnswersController.scala
BabuRupoll7623: @kberec @B-Savage When clicks on the "CHANGE" link associated with the "TAX RETURN" on the Basic CYA (Check Your Answers) page. then user is navigated to the "Check-Dates" page, where they are presented with an option(YES,NO).
If the user selects "YES":
* Any existing data related to the accounting period will be cleaned up (i.e., removed or reset).
* Next Step: After the cleanup, the user will be directed back to the Basic CYA page.
If the user selects "NO":
* Check for Existing Accounting Period Data:
    * If there is existing accounting period data:
        * Next Step: The user will be directed to the accounting period CYA page (Check Your Answers for the accounting period).
    * If there is no existing accounting period data:
        * Next Step: The user will be directed to the Accounting Period Question page.
        
I will update in ticket as well

kberec (approved):

PR #833, opened 2024-08-22
PSR-1337 added redirection to loans and moneyborrowed

narendrapaduchuri:


test/controllers/nonsipp/loansmadeoroutstanding/LoansListControllerSpec.scala
```diff
@@ -213,21 +213,30 @@ class LoansListControllerSpec extends ControllerBaseSpec {
...
+                List(),
+                schemeName,
+                viewOnlyViewModel = Some(viewOnlyViewModel.copy(viewOnlyUpdated = true))
+              )
             )
-          )
       }.withName("OnPageLoadViewOnly renders ok with no loans")
     )
 
+    act.like(
```
kberec, line 232: Could you please move this from "LoansListController in view only mode" to "LoansListController"? As it is not part of the view only mode.
Same for borrowing

kberec (approved):

PR #838, opened 2024-08-30
PSR-1322 change and view declaration links

narendrapaduchuri:


kberec:
As of 17.09.2024, this is blocked by PSR-1373

kberec:
@narendrapaduchuri please check if we can do this in the next sprint.

narendrapaduchuri:
Hi @kberec , I have to check with QA on this, then yes we can pick for next sprint.

PR #845, opened 2024-09-09
PSR1141 group content

narendrapaduchuri:


app/views/templates/Layout.scala.html
```diff
@@ -59,6 +59,7 @@
 
 @additionalScripts = {
     <script src='@controllers.routes.Assets.versioned("javascripts/application.min.js")' @CSPNonce.attr></script>
+    <script @{CSPNonce.attr} src="@routes.Assets.versioned("javascripts/print.js")"></script>
```
kberec, line 62: This will affect every page. Is there a way to add this only to the pages where it is needed? e.g. return submitted page in ticket?

narendrapaduchuri: Need to check, @kberec

kberec (approved):

PR #855, opened 2024-09-16
Content update - Change to hint text

charlotteothon:
Page: What was the total cost of the land or property, in pounds and pence? Change to hint text.


joshhumphreys (commented):
Approved from UCD side

kberec (approved):

PR #856, opened 2024-09-17
PSR-1320 Fixed hard delete issue

westwater:


app/transformations/MemberPaymentsTransformer.scala
```diff
@@ -488,21 +488,43 @@ class MemberPaymentsTransformer @Inject()(
...
+                  logger.info(s"[identifyNewMembers] Member at index $index is changed - NOT safe to hard delete")
+                  None
+                case (index, currentMemberDetail)
+                    if currentMemberDetail.state._new
+                      && psrStatus.exists(_.isSubmitted)
+                      && currentMemberDetail.memberPSRVersion.exists(psrVersion.contains) =>
+                  logger.info(
+                    s"[identifyNewMembers] Member at index $index is new but PSR status is Submitted - NOT safe to hard delete"
+                  )
+                  None
```
kberec, line 511: It looks this case and the below one are not covered with unit tests.

app/transformations/MemberPaymentsTransformer.scala
```diff
@@ -488,21 +488,43 @@ class MemberPaymentsTransformer @Inject()(
...
+                case (index, currentMemberDetail)
+                    if currentMemberDetail.state._new
+                      && psrStatus.exists(_.isCompiled)
+                      && currentMemberDetail.memberPSRVersion.exists(version => !psrVersion.contains(version)) =>
+                  logger.info(
+                    s"[identifyNewMembers] Member at index $index is new and PSR status is Compiled but member PSR version is NOT the same as record PSR version - NOT safe to hard delete"
+                  )
+                  None
```
kberec, line 519: It looks this case and the above one are not covered with unit tests.

westwater: This one is covered by an existing test `should NOT set SafeToHardDelete when PSR status is Compiled but member version is not the same as PSR version and MemberState is New`

app/transformations/MemberPaymentsTransformer.scala
westwater: this is pre-existing logic that got moved about due to formatting - the test is `should NOT set SafeToHardDelete when member and PSR version are the same and greater than 001 and MemberStatus is New -" +
          "this member was added in the previous version but since the state is Submitted," +
          "a change must first be sent to ETMP to update the version to current + 1 and state to Compiled`

kberec (dismissed):

kberec (approved):

PR #866, opened 2024-09-23
PSR-1429 row 2 will now be ignored

narendrapaduchuri:


app/services/MemberDetailsUploadValidator.scala
```diff
@@ -97,21 +95,19 @@ class MemberDetailsUploadValidator @Inject()(
...
-        .drop(1) // drop csv header and process rows
+        .drop(2) // drop csv header + first example row
         .statefulMap[UploadState, Upload](() => UploadState.init)(
           (state, bs) => {
             counter.incrementAndGet()
             if (state.row > Constants.maxSchemeMembers) {
               state.next() -> UploadMaxRowsError
             } else {
-              val parts = bs.map(_.utf8String)
+              val parts = bs.map(_.utf8String).drop(1) // drops the first column
```
kberec, line 105: Why do we need to drop a column here @cyeniceri ?

kberec (approved):

PR #867, opened 2024-09-23
Content update - change to member template file download text

charlotteothon:
Added text: "or guidance text row (row 2),"


joshhumphreys (commented):
Approved from UCD side, ready for dev review

PR #873, opened 2024-09-25
Content change - button text from 'continue' to 'submit'

charlotteothon:
Content update to declaration button text: changed from 'Agree and continue' to 'Agree and submit'


joshhumphreys (commented):
Approved from UCD side

conf/messages.en
```diff
@@ -16,7 +16,7 @@ site.start = Start
 
 site.continue = Continue
 site.saveAndContinue = Save and continue
-site.agreeAndContinue = Agree and continue
+site.agreeAndContinue = Agree and submit
```
kberec, line 19: Do you need to do the same change [here](https://github.com/hmrc/pension-scheme-return-frontend/blob/5f69fc5d2e3af62213d8dee1744ddc54a38ed541/conf/messages.en#L339) and [here](https://github.com/hmrc/pension-scheme-return-frontend/blob/5f69fc5d2e3af62213d8dee1744ddc54a38ed541/conf/messages.en#L355) too?

PR #874, opened 2024-09-25
Content change - wording of important message on declaration page

charlotteothon:
Changed: 'The number of scheme members has changed and is now over 100' to 'The number of scheme members has changed and is now 99 or more'


kberec:
replaced with other PR

PR #889, opened 2024-10-01
PSR-1482_Pension-Payments-Fix | Amend Pension Payments status

ccatosdevelopment:


app/controllers/nonsipp/memberpensionpayments/RemovePensionPaymentsController.scala
```diff
@@ -104,6 +108,7 @@ class RemovePensionPaymentsController @Inject()(
                   .fromTry(
                     request.userAnswers
                       .remove(TotalAmountPensionPaymentsPage(srn, index))
+                      .set(PensionPaymentsJourneyStatus(srn), SectionStatus.InProgress)
```
kberec, line 111: Would this cause that the pension payments task list to be "in progress" if you soft delete any member?

ccatosdevelopment: Good question, thanks for asking! I went and doubled checked our logic and the answer is no, because the status for Pension Payments is calculated using the Yes/No question, combined with the number of members who have `TotalAmountPensionPaymentsPage` set, so `PensionPaymentsJourneyStatus` doesn't directly affect the status within a given session.

It _is_ taken into account by our transformation code, so I traced through what would happen if 1 or more pension payments are left after this removal and `PensionPaymentsJourneyStatus` is `InProgress`, and the "X payments" status would be retained even after logging out and back in, because the Yes answer and number of members with `TotalAmountPensionPaymentsPage` set will be retained.

In the scenario where 0 pension payments are left after this removal, the status would change to `In progress`, and this would be retained after logging in and out, which is the expected behaviour for that scenario - so from what I can see, this should be correct on all counts.

kberec (approved):

PR #890, opened 2024-10-02
PSR-1482 TODO CheckMemberDetailsFileController + errors + tests

Arei1997:


app/controllers/nonsipp/memberdetails/CheckMemberDetailsFileController.scala
```diff
@@ -63,26 +63,39 @@ class CheckMemberDetailsFileController @Inject()(
...
+          case Some(uploadStatus) =>
+            uploadStatus match {
+              case success: UploadStatus.Success =>
+                auditUpload(srn, success, startTime)
+                Ok(view(preparedForm, viewModel(srn, Some(success.name), mode)))
+
+              case failed: UploadStatus.Failed =>
+                auditUpload(srn, failed, startTime)
+                val errorForm = form.withError("file", "uploadMemberDetails.error.failed")
+                Ok(view(errorForm, viewModel(srn, None, mode)))
```
kberec, line 87: @Arei1997 this will keep refreshing the page if the failure happens. One way to mitigate this would be to redirect to a different page and show an error there, as suggested on slack conversation. Might me a bit tricky how to force the error to be displayed properly.
Another way would be to extend viewModel with an additional parameter e.g. "doNotRefresh: Boolean = true" and make the refresh in viewModel conditional on that parameter too. May need to hide the radio on UI (or display a different view) and on button continue to `/upload-member-details-file`.

kberec (approved):
Looks good to me. Needs commits squashed and re-testing.

conf/messages.en
```diff
@@ -3710,3 +3710,5 @@ assetDisposal.reportedOtherAssetsDisposalList.view.none.value = No
 removeAssetDisposal.title = Are you sure you want to remove details of the disposal of this asset?
 removeAssetDisposal.heading = Are you sure you want to remove details of the disposal of {0}?
 removeAssetDisposal.error.required = Select yes if you are sure you want to remove this asset disposal
+
+uploadMemberDetails.error.failed = The file upload failed. Please try again.
```
kberec, line 3714: It is a very minor thing, but if you are squashing commits anyway... could you please move this up in the same file, where the other uploadMemberDetails messages are?

Arei1997: Done 👍

PR #891, opened 2024-10-02
PSR-885 Changed mappings for land or property manual input

kberec:


app/controllers/nonsipp/landorproperty/LandPropertyAddressManualController.scala
```diff
@@ -199,13 +199,19 @@ object LandPropertyAddressManualController {
       "landPropertyAddressManual.field5.error.max"
     )
 
+  private val postCodeFormErrors = InputFormErrors.postcode(
+    "landOrPropertyPostcodeLookup.postcode.error.required",
```
ccatosdevelopment, line 203: Would it be worth replacing this with an empty string, like with fields 2 and 3 on lines 176 and 183? Or is there a particular reason why this is here but the others have been replaced with empty strings?

ccatosdevelopment, line 203: Would it be worth replacing this with an empty string, like with fields 2 and 3 on lines 176 and 183? Or is there a particular reason why this is here but the others have been replaced with empty strings?

The same is true for "landPropertyAddressManual.field5.error.required" on line 197.

app/controllers/nonsipp/landorproperty/LandPropertyAddressManualController.scala
```diff
@@ -199,13 +199,19 @@ object LandPropertyAddressManualController {
       "landPropertyAddressManual.field5.error.max"
```
ccatosdevelopment, line 199: I appreciate this wasn't implemented by you, but I noticed that the messages.en line this relates to is "Enter a postcode in the right format, like AB1 1AB", which seems incorrect. Perhaps something like "Postal code must be 35 characters or fewer" would be better?

ccatosdevelopment (approved):

app/controllers/nonsipp/landorproperty/LandPropertyAddressManualController.scala
kberec: "Enter a postcode in the right format, like AB1 1AB" is how it is in the content document attached to the original story and it is a similar text for the validation message on the postcode lookup page. If you think it could be improved we may need to discuss with our content designer. For now I'd keep it.

app/controllers/nonsipp/landorproperty/LandPropertyAddressManualController.scala
kberec: Yes, you are right this is unused, could be removed. As the PR is already merged, I'm going to add this to my next change. Reason for field5 being there is that is was still in the content document and messages file (removing it from there to). For postcode - copy/paste :)

kberec: Removed them here: https://github.com/hmrc/pension-scheme-return-frontend/pull/897/commits/4d5cc2ba59271483e0b3ca3e7299103ec642139d FYI @ccatosdevelopment

PR #895, opened 2024-10-04
PSR-1373 selecting the max report version fix

cyeniceri:


BabuRupoll7623 (approved):

app/controllers/nonsipp/TaskListController.scala
```diff
@@ -69,7 +69,7 @@ class TaskListController @Inject()(
         } else {
           userAnswersUnchangedAllSections(
             request.userAnswers,
-            if (isSubmitted(response.head)) {
+            if (isSubmitted(response.maxBy(_.reportVersion))) {
```
kberec, line 72: Do we need a test to simulate the wrong order?

kberec, line 72: Is there any other place in the code where "wrong" order returned by getVersions would cause a problem?
E.g. `.head` or similar access assuming a sorted order happens.

cyeniceri: Do we really need to test maxBy here ? Instead of making assumption head or last, we're always selecting record with the greatest report version. `head` was okay for stub data (it's reverse ordered) in local or staging but etmp sending in ascending order.

cyeniceri: In SchemeDetailNavigationUtils, we are utilising getVersions as well. And we are also taking the greatest version number here 

```
                // If so, then determine if the latest submitted return from last year was a full return
                val latestVersionNumber = "%03d".format(psrVersionsResponses.map(_.reportVersion).max)
```

cyeniceri: ReturnsSubmittedController, Testing here is worth IMO. reverse might be extra effort to make stub happy. Also this will cause reverse ordering in QA

```
            val listRetHistorySummary = seqPsrVersionsResponse
              .sortBy(_.reportVersion)
              .reverse
```

cyeniceri: In ViewOnlyReturnSubmittedController, we are not applying any ordering and populating in html directly. This also needs testing if order is important.

```
        psrVersionsResponse <- psrVersionsService.getVersions(request.schemeDetails.pstr, year, srn)
        viewModel = ViewOnlyReturnSubmittedController.viewModel(
          srn,
          request.schemeDetails.schemeName,
          retrievedUserAnswers,
          psrVersionsResponse,
          version,
          dashboardLink
        )
```

kberec: I meant in "TaskListControllerSpec", do we need a test where a wrong order is mocked out and to check if pureUserAnswers used for comparison.

kberec: If you think any of these can cause similar issues on QA, could you please ask our testers to check these?

kberec (approved):

PR #906, opened 2024-10-09
PSR-1485 added mis-match description

narendrapaduchuri:


ccatosdevelopment (approved):

kberec (changes requested):
We're looking at the a11y issue introduced by this change.

PR #907, opened 2024-10-09
Psr 1464

cyeniceri:


app/controllers/nonsipp/membercontributions/MemberContributionListController.scala
```diff
@@ -343,26 +255,22 @@ object MemberContributionListController {
...
-    val normalModeMessage =
-      if (mode == NormalMode)
-        Option(
-          ParagraphMessage(
-            "ReportContribution.MemberList.paragraph1"
-          ) ++
-            ParagraphMessage(
-              "ReportContribution.MemberList.paragraph2"
-            )
+    val optDescription = Option.when(mode == NormalMode)(
```
kberec, line 258: Is this now changes the behaviour?
I..e previously it evaluated to `Option(ParagraphMessage("")` if not NormalMode, now it evaluates to `None`

cyeniceri: https://github.com/hmrc/pension-scheme-return-frontend/blob/main/app/views/TwoColumnsTripleAction.scala.html#L106

I didn't see any behaviour change on my local. As it's optional, it's cleaner to send None instead of Empty String if not NormalMode unless I miss any other aspects.

kberec: Thanks @cyeniceri !  
@Vamshi-Muniganti could you please double-check if these descriptions are still working as expected when testing this ticket?

app/viewmodels/models/ActionTableViewModel.scala
```diff
@@ -46,12 +45,7 @@ case class ActionTableViewModel(
   inset: DisplayMessage,
   head: Option[List[TableElem]],
   rows: List[List[TableElem]],
-  radioText: Message,
-  // whether to render the radio buttons to add another entity to the list or continue
-  showRadios: Boolean = true,
```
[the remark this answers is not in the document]
cyeniceri: It is still in used in ListViewModel. So existing LandOrPropertyListController and it's UI has already radioText/showRadios. ActionTableViewModel is only used in MemberPayments pages and ReturnSubmitted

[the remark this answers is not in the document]
kberec: Yes, we need to find out if member payment list pages are going to have a similar radio button.

kberec (approved):

kberec (approved):

PR #926, opened 2024-10-18
PSR-1510 test call of pre-populated POST endpoint. Made landOrPropert…

kberec:
…yHeld and isLandOrPropertyResidential optional


kberec:
Replaced with https://github.com/hmrc/pension-scheme-return-frontend/pull/944

PR #927, opened 2024-10-20
PSR-1510 copy over of data from previous TY and clear out of details

cyeniceri:


app/controllers/OverviewController.scala
```diff
@@ -233,29 +253,43 @@ class OverviewController @Inject()(
...
+          case _ =>
+            val yearFrom = LocalDate.parse(taxYear)
+            val yearTo = yearFrom.plusYears(1).minusDays(1)
+            val dateRange = DateRange(yearFrom, yearTo)
+            for {
+              userAnswers <- Future.fromTry(request.userAnswers.set(WhichTaxYearPage(srn), dateRange))
+              _ <- saveService.save(userAnswers.copy(id = UNCHANGED_SESSION_PREFIX + userAnswers.id))
+              _ <- saveService.save(userAnswers)
+            } yield {
+              Redirect(controllers.routes.WhatYouWillNeedController.onPageLoad(srn, "", taxYear, version))
```
kberec, line 289: We might need to conditionally redirect to a different URL here in case of pre-population. Do we have the information that the pre-population has just happened?

app/controllers/OverviewController.scala
```diff
@@ -47,8 +49,8 @@ class OverviewController @Inject()(
   config: FrontendAppConfig,
   identify: IdentifierAction,
   allowAccess: AllowAccessActionProvider,
-  getData: DataRetrievalAction,
   createData: DataCreationAction,
+  prePopData: PrePopDataAction,
```
kberec, line 53: Just to avoid confusion, would it be possible not to abbreviate this? I.e. use `prePopulatedData: PrePopulationDataAction` or similar here...

app/controllers/OverviewController.scala
cyeniceri: We can understand it if `lastSubmittedPsrFbInPreviousYears` is not None. That means we are going to read prepop data from etmp.

I separated the logic from SIPP. Do you think it is correct ?

app/controllers/OverviewController.scala
```diff
@@ -110,10 +112,28 @@ class OverviewController @Inject()(
                 messages("site.continue")
               )
             } else {
+              val lastSubmittedPsrFbInPreviousYears =
+                versionsForYears
+                  .filter(x => LocalDate.parse(x.startDate) < yearFrom)
```
kberec, line 117: Would it be better to explicitly search for current tax year - 1  here? In that case the located data would be always from the previous year

app/controllers/OverviewController.scala
kberec: I think it is ok If we can conditionally redirect and use this later on in the UI flow to distinguish pro-population.

app/controllers/OverviewController.scala
cyeniceri: yeah but what happens if we have a gap year ? That's why I used this logic. I remember we asked this to Josh and he said this is not clear however the most recent submission must be used for the population. lastSubmittedPsrFbInPreviousYears logic must guarantee it.

kberec: I've asked this question again, let's see if this is more clear now from UX perspective.

app/controllers/OverviewController.scala
```diff
@@ -233,29 +253,43 @@ class OverviewController @Inject()(
...
+                .addingToSession(Constants.TAX_YEAR -> taxYear)
+                .addingToSession(Constants.VERSION -> version)
+            )
+          case _ =>
+            val yearFrom = LocalDate.parse(taxYear)
+            val yearTo = yearFrom.plusYears(1).minusDays(1)
+            val dateRange = DateRange(yearFrom, yearTo)
+            for {
+              userAnswers <- Future.fromTry(request.userAnswers.set(WhichTaxYearPage(srn), dateRange))
+              _ <- saveService.save(userAnswers.copy(id = UNCHANGED_SESSION_PREFIX + userAnswers.id))
```
kberec, line 286: Maybe I misunderstood something, but I'm struggling to find where does `request.prePopUserAnswers` got copied over to request.userAnswers. Should this happen here?

app/controllers/actions/PrePopDataAction.scala
```diff
@@ -0,0 +1,76 @@
...
+              prePopReturn <- psrRetrievalService.getAndTransformStandardPsrDetails(
+                optFbNumber = Some(lastSubmittedPsrFbInPreviousYears),
+                fallBackCall = controllers.routes.OverviewController.onPageLoad(existingDataRequest.srn)
+              )(
+                hc = implicitly,
+                ec = implicitly,
+                request = DataRequest[A](allowedAccessRequest, emptyUserAnswers(allowedAccessRequest))
+              )
+              cleanedPrePopUa = cleanUpPrePop(prePopReturn)
+              _ <- sessionRepository.set(cleanedPrePopUa.copy(id = PREPOP_PREFIX + prePopReturn.id))
```
kberec, line 56: Should the prePopulated UA get deleted somewhere. Now if you navigate back to the overview page on the UI and go down on the View and change route, request.prePopUserAnswers is still there.

app/controllers/OverviewController.scala
cyeniceri: It's not copied over yet. After cleaning up either we can copy `request.prePopUserAnswers` into currentUA and (delete `request.prePopUserAnswers` or keep it here )

app/controllers/actions/PrePopDataAction.scala
cyeniceri: If I consider this question and above one together, we may copy and paste over current UA and delete the prepopUA in PrePopulationDataAction. What do you think ?

app/controllers/OverviewController.scala
kberec: There is some ongoing discussion about this, we're going to keep this as is for now until that is resolved.
Could you please add a comment @cyeniceri so we can easily identify this later on, in case we need to changed it? E.g. something like `\\ this logic assumes gap years are allowed in pre-population`

app/controllers/actions/PrePopDataAction.scala
kberec: I'm not sure. If we are deleting it immediately, what is the point of having `request.prePopulationUserAnswers`? Cannot we just populate request.userAnswers directly?

kberec: We agreed to copy over `request.prePopulationUserAnswers` to `request.userAnswers` and delete `request.prePopulationUserAnswers` after the cleanup (deletion of data that not needed in the new return).

app/controllers/OverviewController.scala
kberec: We discussed with @cyeniceri and agreed to have a flag saved somewhere (e.g. request or mongo) to indicate that this return is a pre-population.
The value of this flag would be populated on "Start" and on "Continue" link too.
This value could be re-used later on for a conditional navigation and/or the task list/cya pages to dictate if this is a pre-population. FYI @westwater @ccatosdevelopment @narendrapaduchuri .

kberec: Related: https://github.com/hmrc/pension-scheme-return-frontend/pull/930#discussion_r1810634372

kberec:
@cyeniceri I'm closing as it is replaced by https://github.com/hmrc/pension-scheme-return-frontend/pull/946, please check if this was right

PR #930, opened 2024-10-22
PSR-1510 list page

westwater:


app/viewmodels/models/ListViewModel.scala
```diff
@@ -42,33 +45,55 @@ object ListRow {
...
-      view = Some(ListRowNoLink(value))
+      view = Some(ListRowNoLink(value)),
+      check = None
     )
 }
 
 case class ListViewModel(
   inset: DisplayMessage,
-  rows: List[ListRow],
+  sections: List[Section],
```
westwater, line 82: main change - list pages now have a list of sections instead of rows, each section is just a heading and a list of rows. This can be expanded to house section specific elements like pagination if needed later on

app/views/ListView.scala.html
```diff
@@ -55,35 +55,45 @@ <h1 class="govuk-heading-l">@renderMessage(viewmodel.heading)</h1>
...
-                            value.change.map(change => ListWithActionsActionViewModel(messages("site.change"), change.url)
-                            .withVisuallyHiddenText(change.hiddenText.toMessage)).toList ++
-                                    value.remove.map(remove => ListWithActionsActionViewModel(messages("site.remove"), remove.url)
-                                    .withVisuallyHiddenText(remove.hiddenText.toMessage)).toList ++
-                                    value.view.map {
-                                        case ListRowLink(url, hiddenText) => ListWithActionsActionViewModel(messages("site.view"), url)
-                                        .withVisuallyHiddenText(hiddenText.toMessage)
-                                        case ListRowNoLink(text) => ListWithActionsActionViewModel.noLink(renderMessage(text))
-                                    }.toList
+            @viewmodel.page.sections.map { section =>
```
westwater, line 58: iterate through each section and renders the rows. also optionally render a heading is it exists

app/controllers/nonsipp/landorproperty/LandOrPropertyListController.scala
```diff
@@ -102,21 +102,37 @@ class LandOrPropertyListController @Inject()(
...
     if (status == TaskListStatus.NotStarted) {
       Redirect(routes.LandOrPropertyHeldController.onPageLoad(srn, NormalMode))
     } else if (status == TaskListStatus.InProgress) {
       Redirect(incompleteLandOrPropertyUrl)
     } else {
-      val addresses = userAnswers.map(LandOrPropertyAddressLookupPages(srn))
+
+      // Dependant on Christo's work
+
+      // Get all indexes for a section that have atleast completed the first page
```
kberec, line 122: Is there a need to have a Mode.PrePopulation and task list, list page and CYA pages to load with this mode?
In that case prePopAddresses would have data only if this case of pre-population mode.
To reduce the risk of affecting existing functionality.
If mode is an overkill, we could consider another way of distinguishing pre-population mode, e.g. store a boolean flag `isPrePopulation' in the request.

app/views/ListView.scala.html
```diff
@@ -55,35 +55,45 @@ <h1 class="govuk-heading-l">@renderMessage(viewmodel.heading)</h1>
...
 
+                @defining(viewmodel.page.paginatedViewModel.fold(section.rows)(paginatedViewModel =>
+                    section.rows.paginate(paginatedViewModel.pagination.currentPage, paginatedViewModel.pagination.pageSize))
+                ) { rows =>
+
+                    @section.label.map { label =>
```
kberec, line 64: Label would need to be skipped if there are no items.

app/views/ListView.scala.html
```diff
@@ -26,7 +26,7 @@
 )
 @(form: Form[Boolean], viewmodel: FormPageViewModel[ListViewModel])(implicit request: Request[_], messages: Messages)
 
-    @layout(pageTitle = title(form, if(viewmodel.optViewOnlyDetails.nonEmpty) viewmodel.optViewOnlyDetails.get.title.toMessage else viewmodel.title.toMessage),showBackLink = viewmodel.showBackLink) {
+    @layout(pageTitle = title(form, if(viewmodel.optViewOnlyDetails.nonEmpty) viewmodel.optViewOnlyDetails.get.title.toMessage else viewmodel.title.toMessage), showBackLink = viewmodel.showBackLink) {
```
kberec, line 29: Title (and heading) should be different in pre-population mode. Can be set up by the controller and just passed in as `viewmodel.title` value though.

PR #938, opened 2024-10-29
PSR-1556 | Implement LOP Check Methods

ccatosdevelopment:


app/utils/nonsipp/CheckStatusUtils.scala
```diff
@@ -0,0 +1,175 @@
...
+   * @param srn the Scheme Reference Number, used for the .get calls
+   * @param recordIndex the index of the record being checked
+   * @param identitySubject relates to the section this is called from: Loans, Land or Property, Shares, or Other Assets
+   * @param identityType relates to the seller involved: Individual, UKCompany, UKPartnership, or Other
+   * @return true if all answers are present, else false
+   */
+  def identitySubjectAnswersPresent(
+    userAnswers: UserAnswers,
+    srn: Srn,
+    recordIndex: Max5000,
```
westwater, line 129: are the identity subject sections in other journeys indexed by this type?

app/utils/nonsipp/CheckStatusUtils.scala
```diff
@@ -0,0 +1,175 @@
...
+  /**
+   * This method determines whether or not all answers are present for a given IdentityType & IdentitySubject.
+   * @param userAnswers the answers provided by the user
+   * @param srn the Scheme Reference Number, used for the .get calls
+   * @param recordIndex the index of the record being checked
+   * @param identitySubject relates to the section this is called from: Loans, Land or Property, Shares, or Other Assets
+   * @param identityType relates to the seller involved: Individual, UKCompany, UKPartnership, or Other
+   * @return true if all answers are present, else false
+   */
+  def identitySubjectAnswersPresent(
```
westwater, line 126: very minor: could you mark functions that will not be used outside of this class as private

app/utils/nonsipp/CheckStatusUtils.scala
```diff
@@ -0,0 +1,175 @@
...
+              )
+            }
+          )
+          .contains(true)
+    }
+  }
+
+  /**
+   * This method determines whether or not a Land or Property record needs to be checked. A record needs checking if
+   * all of the pre-populated-then-cleared answers are missing & all of the other answers are present.
```
westwater, line 64: wouldn't the logic need to be
```
A record needs checking if ANY of the pre-populated-then-cleared answers are missing
```

cyeniceri: For LOP, I think after the pre-populated-then-cleared `userAnswers.get(LandOrPropertyHeldPage(srn))` will be empty and that means we need to check here, don't we ?

https://docs.google.com/spreadsheets/d/1ZeT7B6xhryMHjE231WCW_N6MXypo04x0QyzFEKAoraU/edit?pli=1&gid=0#gid=0

app/utils/nonsipp/CheckStatusUtils.scala
ccatosdevelopment: Good question, yes - I just double checked the maximums for Loans, Shares, and Other Assets, and they're all 5000. Bonds also has a limit of 5000, but it doesn't use IdentitySubject, so in any case I think we're good on all fronts.

app/utils/nonsipp/CheckStatusUtils.scala
ccatosdevelopment: That's a good shout, but this method's also used in CheckStatusUtilsSpec, so I can't set it as private because then it'll be inaccessible for those tests.

app/utils/nonsipp/CheckStatusUtils.scala
ccatosdevelopment: I think the rationale behind the 'all' implementation is that we're less likely to show any newly-added-but-incomplete journeys as records that need checking, since the user would have to back out of the journey at exactly the right page for that to happen. If we use 'any' rather than 'all', it's more likely that a newly-added-but-incomplete journey will end up being displayed as needing checking.

The drawback of this is that if a user goes to check a pre-populated journey and only provides partial answers, then backs out, then this journey will no longer be displayed as needing checking. This was deemed unlikely to happen, so the drawbacks of the 'all' implementation were deemed to be less likely to occur than the drawbacks of the 'any' implementation.

Neither implementation is ideal, which is why I thought it would be worth looking into another suggested solution, but that wasn't possible within the given timeframe, so for now the decision was made to take the 'all' approach.

ccatosdevelopment: @cyeniceri That will be checked by the `checkLandOrPropertySection` method when the user lands on the Task List page, whereas this method will be called for each method on the new LOP List page implementation, at which point we already know that the user has some LOP and so `LandOrPropertyHeldPage` must be `true`.

app/utils/nonsipp/CheckStatusUtils.scala
westwater: if you're dead set on testing this function in isolation, you can set it to package private

ccatosdevelopment: Okay, I'll take what you've both said onboard and set it as private, then restructure the tests accordingly. I'll push that as a separate commit, and if the consensus is that it's better that way then we can keep that commit, and if not, then we can revert that commit.

ccatosdevelopment: No it's alright @westwater , I don't feel particularly strongly about it, I just thought it made sense to organise the tests by each method they're testing.

westwater (approved):

PR #943, opened 2024-10-30
Psr 1468 - Disposal journey CYA & Change details page, maximum 50 disposal condition and inset text

Arei1997:


westwater (approved):

app/controllers/nonsipp/sharesdisposal/SharesDisposalCYAController.scala
```diff
@@ -81,148 +82,159 @@ class SharesDisposalCYAController @Inject()(
...
-            viewOnlyUpdated = false, // flag is not displayed on this tier
-            optYear = request.year,
-            optCurrentVersion = request.currentVersion,
-            optPreviousVersion = request.previousVersion,
-            compilationOrSubmissionDate = request.userAnswers.get(CompilationOrSubmissionDatePage(srn))
+        disposalAmount = updatedUserAnswers
+          .map(SharesDisposalProgress.all(srn, shareIndex))
+          .count { case (_, progress) => progress.completed }
+
+        _ <- saveService.save(updatedUserAnswers).liftF
```
kberec, line 195: @Arei1997 @westwater @Vamshi-Muniganti Looks that this save broke the change share disposals functionality by clearing out `SharesDisposalCYAPointOfEntry` value.
Do you by any chance remember why did you need to save again all this data at this point on onPageLoadCommon?

Arei1997: I just had a look

This save was used so that what was shown on the CYA page reflected the "completed" state accurately of the disposal journey before being rendered by the view. It was also to ensure a persisted state, if i remember correctly before this change, disposals were not all represented to then be recognised to show the correct inset test

kberec: Hm, I would think that `SharesDisposalProgress` should be already set before this CYA page is hit.

PR #944, opened 2024-10-31
PSR-1558 PSR-1563

kberec:


app/services/PsrSubmissionService.scala
```diff
@@ -120,6 +116,31 @@ class PsrSubmissionService @Inject()(
       )
   }
 
+  private def callSubmission(
+    srn: Srn,
+    submissionRequest: PsrSubmission,
+    isSubmitted: Boolean
+  )(implicit hc: HeaderCarrier, ec: ExecutionContext, request: DataRequest[_]): Future[Either[String, Unit]] =
+    if (isPrePopulation(implicitly) && !isSubmitted) {
```
cyeniceri, line 124: we don't need to say (implicitly) here explicitly unless we have another parameter implicitly and provide it explicitly. For example :
```
private def isPrePopulation(implicit request: DataRequest[_], int: explicit): Boolean = ???
isPrePopulation(implicitly, 1)
```

but in our case `if (isPrePopulation && !isSubmitted)`  must be alright.

app/transformations/LandOrPropertyTransformer.scala
```diff
@@ -47,19 +48,25 @@ class LandOrPropertyTransformer @Inject() extends Transformer {
   def transformToEtmp(srn: Srn, optLandOrPropertyHeld: Option[Boolean], initialUA: UserAnswers)(
     implicit request: DataRequest[_]
   ): Option[LandOrProperty] =
-    optLandOrPropertyHeld.map(landOrPropertyHeld => {
-      val disposeAnyLandOrProperty = request.userAnswers.get(LandOrPropertyDisposalPage(srn)).getOrElse(false)
+    Option.when(optLandOrPropertyHeld.nonEmpty || request.userAnswers.map(LandPropertyInUKPages(srn)).toList.nonEmpty) {
+      val optDisposeAnyLandOrProperty = request.userAnswers.get(LandOrPropertyDisposalPage(srn))
+      val dispose =
+        if (isPrePopulation(implicitly) && optDisposeAnyLandOrProperty.isEmpty) {
```
cyeniceri, line 54: same here as above.
Should we reuse following method ? This is second time I have seen it.
private def isPrePopulation(implicit request: DataRequest[_]): Boolean =
    request.session.get(PREPOPULATION_FLAG).fold(false)(_.toBoolean)

kberec: I agree, can I tidy that up in my next ticket @cyeniceri ?

cyeniceri (approved):

PR #946, opened 2024-10-31
PSR-1568 Copy over pre-populated data from the previous tax year

cyeniceri:


app/controllers/OverviewController.scala
```diff
@@ -278,6 +309,12 @@ class OverviewController @Inject()(
               .getOrElse(result)
           }
         case _ =>
+          Future.successful(
+            Redirect(controllers.nonsipp.routes.TaskListController.onPageLoad(srn))
+              .addingToSession(
+                Constants.PREPOPULATION_FLAG -> String.valueOf(lastSubmittedPsrFbInPreviousYears.isDefined)
```
kberec, line 315: Will 'onSelectContinue' got called for submissions(version) greater than 1? As PREPOPULATION_FLAG should not be set for version 1 only.

app/controllers/OverviewController.scala
```diff
@@ -233,36 +242,58 @@ class OverviewController @Inject()(
...
+            val yearTo = yearFrom.plusYears(1).minusDays(1)
+            val dateRange = DateRange(yearFrom, yearTo)
+            for {
+              userAnswers <- Future.fromTry(request.userAnswers.set(WhichTaxYearPage(srn), dateRange))
+              _ <- saveService.save(userAnswers.copy(id = UNCHANGED_SESSION_PREFIX + userAnswers.id))
+              _ <- saveService.save(userAnswers)
+            } yield {
+              if (lastSubmittedPsrFbInPreviousYears.isDefined) {
+                Redirect(controllers.routes.WhatYouWillNeedController.onPageLoad(srn, "", taxYear, version))
+                  .addingToSession(Constants.PREPOPULATION_FLAG -> String.valueOf(true))
```
kberec, line 280: @cyeniceri do you think it is safe to store this in the session, as this value will be used quite further down the line? Just thinking about the issues we seen in PB that can be due to the restart of the application (or any other reasons that clear out the session).
Should we store it in mongo FE cache instead? or in request?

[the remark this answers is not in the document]
cyeniceri: That makes sense. We could have decided by doing `request.session.get(PREPOPULATION_FLAG).getOrElse("false")`. But setting false makes it clearer. I did it

app/controllers/OverviewController.scala
cyeniceri: Yes, It will. What should be the value of PREPOPULATION_FLAG for for submissions(version) greater than 1. I thought false made it (`String.valueOf(lastSubmittedPsrFbInPreviousYears.isDefined)`) but I might be wrong

cyeniceri: as I'm sending None now for previous data (versions greater than 1)

kberec: "What should be the value of PREPOPULATION_FLAG for for submissions(version) greater than 1?" - I think it should be false.
It's worth double checking, I think.

kberec (commented):
Looks good. Can be tested. We'd need to wait with the merge a bit.

kberec (approved):

PR #947, opened 2024-11-04
PSR-721 - maximum character lengths that correspond to ETMP bug fixes

Arei1997:


kberec:
@Arei1997 I closed this, as the approach has been changed and a separate PR is already merged.

PR #948, opened 2024-11-04
PSR-629 fixed Other Char issue

narendrapaduchuri:


app/forms/TextFormProvider.scala
```diff
@@ -52,6 +53,21 @@ class TextFormProvider @Inject()() {
     )
   )
 
+  def textAreaOther(
```
kberec, line 56: Would it be ok to refactor/rename this method to `textAreaMinLength`

kberec: hold on that can be misleading too...

kberec (approved):

PR #951, opened 2024-11-06
PSR-1572

cyeniceri:


test/services/PrePopulationServiceSpec.scala
```diff
@@ -167,4 +188,2089 @@ object PrePopulationServiceSpec {
         )
       )
   }
+
+  val baseReturnJsValue: JsValue =
```
kberec, line 192: Is there a better way to construct the base return / clean result value?
My concern is that if we change anything in the UA data structure (e.g. rename a field or move it to somewhere else in the structure) we would need to locate that in this json and manually correct it.

test/services/PrePopulationServiceSpec.scala
```diff
@@ -167,4 +188,2089 @@ object PrePopulationServiceSpec {
...
+      |                "key" : "Sold"
+      |              },
+      |              "1" : {
+      |                "key" : "Sold"
+      |              }
+      |            }
+      |          },
+      |          "portionStillHeld" : {
+      |            "0" : {
+      |              "0" : false
```
kberec, line 549: Do we need to test something like:

```
"0" : true,
"1" : false
```

app/services/PrePopulationService.scala
```diff
@@ -49,4 +60,44 @@ class PrePopulationService @Inject() {
...
+        .as[Map[String, Map[String, Boolean]]]
+
+    val transformedResult = baseUaJson
+      .transform(landOrProperty.json.pickBranch)
+      .flatMap(_.transform(LandOrPropertyRecordVersionPage(srn).path.prune(_)))
+      .flatMap(_.transform(LandOrPropertyHeldPage(srn).path.prune(_)))
+      .flatMap(_.transform(LandOrPropertyDisposalPage(srn).path.prune(_)))
+      .flatMap(_.transform(disposalPropertyTransaction.prune(_)))
+      .flatMap(_.transform((Paths.heldPropertyTransactions \ "isLandOrPropertyResidential").prune(_)))
+      .flatMap(_.transform((Paths.heldPropertyTransactions \ "landOrPropertyLeased").prune(_)))
```
kberec, line 78: It would be possible to add case class `IsLandOrPropertyResidentialPages` (same as `IsLandOrPropertyResidentialPage` but without the last index part in path). That would avoid  defining `"landOrPropertyLeased"` here. 
What do you think, would it be cleaner or it is an overkill?

test/services/PrePopulationServiceSpec.scala
cyeniceri: I'll do

cyeniceri: this new case pushed now

app/services/PrePopulationService.scala
cyeniceri: I removed this 
`      .flatMap(
        _.transform(
          (Paths.heldPropertyTransactions \ "leaseDetails" \ "connectedPartyStatus" \ "isLesseeConnectedParty").prune(_)
        )
      )`
      
This is already covered by 

`.flatMap(_.transform((Paths.heldPropertyTransactions \ "leaseDetails").prune(_)))`

test/services/PrePopulationServiceSpec.scala
cyeniceri: Currently I'm copying this string (json) and pasting to an editor (in my IntelliJ json file) and fix/modify etc..
We can put this data in a json file and load it from there if we think this way will be easier. 

Producing this json data by building a complex scala object by setting case classes seems more complex to me.

What do you think @kberec ?

app/services/PrePopulationService.scala
kberec: That's great. What I meant above is that we can possibly avoid defining e.g. "landOrPropertyLeased" in multiple places: here and in `IsLandOrPropertyResidentialPage`. I.e. if we define it as path in IsLandOrPropertyResidentialPages and use the path here. But it's very minor and might be pros and cons, so resolving.

kberec (approved):

PR #953, opened 2024-11-08
PSR-1322-1 view and change declaration link

narendrapaduchuri:


app/utils/nonsipp/TaskListUtils.scala
```diff
@@ -86,7 +95,37 @@ object TaskListUtils {
...
+        isLinkActive,
+        noChangesSincePreviousVersion,
+        schemeName
+      )
+
+    sectionListWithoutDeclaration :+ declarationSectionViewModel
+
+  }
+
+  def getSectionListOne(
```
ccatosdevelopment, line 123: This method doesn't have any usages, can we delete this?

app/utils/nonsipp/TaskListUtils.scala
```diff
@@ -55,23 +55,32 @@ object TaskListUtils {
...
           }
-        LinkMessage(
-          s"$prefix.complete",
-          psaOrPspDeclarationUrl
-        )
+        }
+        if (noChangesSincePreviousVersion) {
+          LinkMessage(
+            s"$prefix.view",
+            psaOrPspDeclarationUrl
```
ccatosdevelopment, line 76: I think this is supposed to go to the Submission Confirmation page, but it looks like it's still going to the Declaration page - is this correct?

kberec:
Replaced by https://github.com/hmrc/pension-scheme-return-frontend/pull/962

PR #955, opened 2024-11-11
PSR-1571 Additional requestRole header with possible values PSA or PSP

kberec:


test/controllers/nonsipp/shares/SharesListControllerSpec.scala
```diff
@@ -214,6 +215,17 @@ class SharesListControllerSpec extends ControllerBaseSpec {
       }.withName("OnPageLoadViewOnly renders ok with no shares")
     )
 
+    act.like(
```
kberec, line 218: Adding back test that was previously removed by mistake in PSR-1420

westwater (approved):

PR #957, opened 2024-11-12
PSR-1322 Added CheckUpdateInformationPage

kberec:


kberec:
Closed as the branch name was wrong

PR #961, opened 2024-11-13
PSR-1576 Clear out some data from the previous tax year - member paym…

cyeniceri:
…ents

### DIFF
--- a/test/prepop/MemberPrePopulationProcessorSpec.scala
+++ b/test/prepop/MemberPrePopulationProcessorSpec.scala
@@ -1,0 +1,907 @@
+/*
+ * Copyright 2024 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


test/prepop/MemberPrePopulationProcessorSpec.scala
```diff
@@ -0,0 +1,822 @@
...
+        |        "memberDetailsSectionCompleted": {
+        |          "0": {},
+        |          "1": {},
+        |          "2": {},
+        |          "3": {},
+        |          "4": {},
+        |          "5": {}
+        |        },
+        |        "memberStatus": {
+        |          "0": "Changed",
```
kberec, line 115: if the status is deleted, member should not be copied over.

cyeniceri: the records with `memberStatus : Deleted` was already marked as `"soft-deleted"` in baseReturn json in retrieval service and we aren't copying over these soft-deleted member records.

kberec: Do you need to test that soft-deleted members are not copied over?

cyeniceri: I put `"soft-deleted"` into baseReturn and verified that it's not within copied over userAnswer after cleanup.

kberec (approved):

kberec (approved):

PR #964, opened 2024-11-20
PSR-1604 | AmountOfTheLoanPage optional values

ccatosdevelopment:


app/pages/nonsipp/loansmadeoroutstanding/AmountOfTheLoanPage.scala
```diff
@@ -21,9 +21,9 @@ import pages.QuestionPage
 import config.RefinedTypes.Max5000
 import models.SchemeId.Srn
 import play.api.libs.json.JsPath
-import models.Money
+import models.AmountOfTheLoan
 
-case class AmountOfTheLoanPage(srn: Srn, index: Max5000) extends QuestionPage[(Money, Money, Money)] {
+case class AmountOfTheLoanPage(srn: Srn, index: Max5000) extends QuestionPage[AmountOfTheLoan] {
```
westwater, line 26: This has caused a lot of `.get`s later on, would it be better if we kept this case class as it was and create a new one with `AmountOfLoan` as the type is just needed for the get on the AmountOdLoanController

westwater: yup thats right. It's personal opinion but I'd usually go for an auxiliary function in a companion object of AmountOfTheLoanPage like def partial(???) only to limit the class space for searching / more intuitive to use etc but thats just me 👍

westwater (approved):

PR #966, opened 2024-11-21
PSR-958 - Inset message on completed disposals

Arei1997:


app/controllers/nonsipp/landorpropertydisposal/LandOrPropertyDisposalListController.scala
```diff
@@ -210,7 +219,8 @@ class LandOrPropertyDisposalListController @Inject()(
                           request.userAnswers,
                           request.schemeDetails.schemeName,
                           showBackLink = true,
-                          maximumDisposalsReached = maximumDisposalsReached
+                          maximumDisposalsReached = maximumDisposalsReached,
+                          allPropertiesFullyDisposed = true
```
kberec, line 223: Is this intentional? Should it be allPropertiesFullyDisposed = allPropertiesFullyDisposed?

Arei1997: That is correct, it should be allPropertiesFullyDisposed = allPropertiesFullyDisposed, I have overlooked that.  As allPropertiesFullyDisposed was not within the onSubmit scope

THis should fix the validation issue

PR #971, opened 2024-11-26
PSR-1578 Clear out some data from the previous tax year - shares

cyeniceri:


app/pages/nonsipp/shares/package.scala
```diff
@@ -50,7 +52,11 @@ package object shares {
       TypeOfSharesHeldPage(srn, index),
       WhenDidSchemeAcquireSharesPage(srn, index),
       WhyDoesSchemeHoldSharesPage(srn, index),
-      SharesCompleted(srn, index)
+      SharesCompleted(srn, index),
+      IdentityTypePage(srn, index, SharesSeller),
```
kberec, line 56: Adding these pages may cause some problems that is hard to discover and result in bugs later on.
If this is for removal, cleanup method in `WhyDoesSchemeHoldSharesPage` should already include these pages.

cyeniceri: It was not clearing those fields without adding here but let me double check it

kberec: Maybe because it uses `removeOnly` and that doesn't call cleanup. If that is the case, just double check please that this doesn't break anything in `RemoveSharesController`.

cyeniceri: https://github.com/hmrc/pension-scheme-return-frontend/blob/ffc50c00147496b536ee71999ea3a10dc05c4c50/app/controllers/nonsipp/shares/RemoveSharesController.scala#L87
It also uses only "remove" method so do I. I think it was already buggy. But yeah will check

BabuRupoll7623: @kberec i have tested in branch and main level with @cyeniceri , CRN ,Other recipient , Partnership details cleanse in PSR-1578 branch ,which not removed in main branch , the existing "remove shares" functionality remains unaffected and is not causing any issues as a result of these changes.

kberec (approved):

ccatosdevelopment (approved):

PR #974, opened 2024-11-27
NO-JIRA Fix flaky test

westwater:
Row value has to be unique

### DIFF
--- a/test-utils/generators/ViewModelGenerators.scala
+++ b/test-utils/generators/ViewModelGenerators.scala
@@ -29,7 +29,7 @@ trait ViewModelGenerators extends BasicGenerators {
 
   def formPageViewModelGen[A](implicit gen: Gen[A]): Gen[FormPageViewModel[A]] =
     for {
-      title <- nonEmptyMessage
+      title <- nonEmptyMessage(30)
       heading <- nonEmptyInlineMessage
       description <- Gen.option(nonEmptyBlockMessage)
       page <- gen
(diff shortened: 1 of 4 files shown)


test-utils/generators/ViewModelGenerators.scala
```diff
@@ -207,23 +207,23 @@ trait ViewModelGenerators extends BasicGenerators {
...
     } yield {
       FurtherDetailsViewModel(title, contents)
     }
 
   def radioListRowViewModelGen: Gen[RadioListRowViewModel] =
     for {
-      content <- nonEmptyMessage
-      value <- nonEmptyString
+      content <- nonEmptyMessage(30)
+      value <- uniqueStringGen
```
westwater, line 219: this is the fix

test-utils/generators/ViewModelGenerators.scala
```diff
@@ -207,23 +207,23 @@ trait ViewModelGenerators extends BasicGenerators {
...
       contents <- nonEmptyDisplayMessage
     } yield {
       FurtherDetailsViewModel(title, contents)
     }
 
   def radioListRowViewModelGen: Gen[RadioListRowViewModel] =
     for {
-      content <- nonEmptyMessage
-      value <- nonEmptyString
+      content <- nonEmptyMessage(30)
```
westwater, line 218: trimmed down the generated string values to make them more digestable when debugging

PR #981, opened 2024-11-29
PSR-1577 Clear out some data from the previous tax year - loans

cyeniceri:


cyeniceri:
Hi @Neozxc , thanks for your review. 
I'm not using my Intellij default formatter, I use scalaFmt file within our project and this formatter is being used among all devs in team. As you suggest, I always format my files. Even I don't apply it manual, when you do `sbt compile test ..` it will format it automatically for you. And I do paying attention to execute `runtests.sh` always to make sure nothing is broken before push.

Neozxc:
> Hi @Neozxc , thanks for your review. I'm not using my Intellij default formatter, I use scalaFmt file within our project and this formatter is being used among all devs in team. As you suggest, I always format my files. Even I don't apply it manual, when you do `sbt compile test ..` it will format files automatically for you. And I do pay attention to execute `runtests.sh` always to make sure nothing is broken before push.

Thanks @cyeniceri. We can leave the formatting out please. But i think we can address the import comments

Neozxc (approved):

PR #982, opened 2024-12-02
PSR-1661 Fixed pagination

westwater:


app/views/ListView.scala.html
```diff
@@ -55,8 +55,12 @@ <h1 class="govuk-heading-l">@renderMessage(viewmodel.heading)</h1>
                 <h1 class="govuk-heading-l">@renderMessage(viewmodel.optViewOnlyDetails.get.heading)</h1>
             }
 
+            @if(viewmodel.page.sections.size == 1){
```
kberec, line 58: Probs this is not needed, please check

.gitignore
```diff
@@ -35,3 +35,8 @@ bin
 *.iws
 /.bsp/sbt.json
 /.java-version
+.vscode/
```
kberec, line 38: Are these needed?

westwater: yeah trying out vscode

app/views/ListView.scala.html
westwater: woops leftover from initial impl

kberec (approved):

PR #987, opened 2024-12-03
PSR-1608 Provide path for either view and change or standard journey …

hmrc-spifix:
…on continue

### DIFF
--- a/app/controllers/OverviewController.scala
+++ b/app/controllers/OverviewController.scala
@@ -34,11 +34,12 @@ import utils.nonsipp.SchemeDetailNavigationUtils
 import models.backend.responses._
 import utils.DateTimeUtils.localDateShow
 import models._
+import play.api.Logger
 import play.api.i18n.{I18nSupport, Messages, MessagesApi}
 import viewmodels.OverviewSummary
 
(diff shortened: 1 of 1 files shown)


app/controllers/OverviewController.scala
```diff
@@ -298,16 +301,29 @@ class OverviewController @Inject()(
...
       reportType match {
         case PsrReportType.Sipp.name =>
-          val sippUrl = s"${config.urls.sippBaseUrl}/${srn.value}${config.urls.sippContinueJourney}"
-          Future.successful {
-            val result = Redirect(sippUrl)
-              .addingToSession(Constants.TAX_YEAR -> taxYear)
-              .addingToSession(Constants.VERSION -> version)
+          version.toIntOption match {
+            case None =>
+              logger.error(s"Could not parse version [$version] to int")
```
kberec, line 306: Could you please add something to indicate this error message is for SIPP? This is so if we see a pager duty alert, to know that this relates to SIPP.

kberec (changes requested):

westwater (approved):

tolgatunaaa (approved):

PR #999, opened 2024-12-06
PSR-1609 | Implement breadcrumbs

ccatosdevelopment:


test/controllers/CheckUpdateInformationControllerSpec.scala
```diff
@@ -31,18 +31,46 @@ class CheckUpdateInformationControllerSpec extends ControllerBaseSpec {
   lazy val onPageLoad: String = routes.CheckUpdateInformationController.onPageLoad(srn).url
   lazy val onSubmit: String = routes.CheckUpdateInformationController.onSubmit(srn).url
 
+  private val dashboardUrlPsa =
+    "http://localhost:8204/manage-pension-schemes/pension-scheme-summary/" + srn.value
```
Neozxc, line 35: Could we make use of interpolation here please? Something like this:

```scala
private val dashboardUrlPsa = s"http://localhost:8204/manage-pension-schemes/pension-scheme-summary/${srn.value}"
```

test/controllers/CheckUpdateInformationControllerSpec.scala
```diff
@@ -31,18 +31,46 @@ class CheckUpdateInformationControllerSpec extends ControllerBaseSpec {
...
+
+  private val dashboardUrlPsp =
```
Neozxc, line 37: Could we also make use of interpolation here please?

test/controllers/WhatYouWillNeedControllerSpec.scala
```diff
@@ -43,18 +43,46 @@ class WhatYouWillNeedPageControllerSpec extends ControllerBaseSpec {
   override protected def beforeAll(): Unit =
     reset(mockPsrRetrievalService)
 
+  private val dashboardUrlPsa =
```
Neozxc, line 46: Could we also make use of interpolation here please?

test/controllers/WhatYouWillNeedControllerSpec.scala
```diff
@@ -43,18 +43,46 @@ class WhatYouWillNeedPageControllerSpec extends ControllerBaseSpec {
...
+    "http://localhost:8204/manage-pension-schemes/pension-scheme-summary/" + srn.value
+
+  private val dashboardUrlPsp =
```
Neozxc, line 49: Could we also make use of interpolation here please?

Neozxc (approved):

PR #1002, opened 2024-12-10
PSR-1678 | [Accessibility] Issue: Incorrect error summary link destination #8763

Neozxc:
**Issue:**
- [Incorrect error summary link destination](https://github.com/hmrc/accessibility-audits-external/issues/8763)

**Dev Tasks:**
- [x] Map ids to the expected field

### DIFF
--- a/app/views/MultipleQuestionView.scala.html
+++ b/app/views/MultipleQuestionView.scala.html
@@ -37,8 +37,8 @@
     @if(form.hasErrors) {
       @govukErrorSummary(ErrorSummaryViewModel(
         form,
-        errorLinkOverrides = Map("value.1" -> "value.1.day")))
-      )
+        errorLinkOverrides = errorLinkOverrideHelper
+      ))
     }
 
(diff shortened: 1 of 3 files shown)


app/views/MultipleQuestionView.scala.html
```diff
@@ -154,3 +157,10 @@ <h1 class="govuk-heading-l govuk-!-margin-bottom-3">@renderMessage(viewModel.hea
         )
     )).withFormField(form(s"value.${index + 1}")))
 }
+
+@errorLinkOverrideHelper = @{
+  viewModel.page.fields.zipWithIndex.collect {
+    case (field @ QuestionField(_, _, _, _, FieldType.Date), index) =>
```
kberec, line 163: Do you need a default case here to cater for anything that is not QuestionField with FieldType.Date?

kberec: I think no, because this is an override, for other type of fields the error links doesn't need an overide.

kberec (approved):

PR #1010, opened 2024-12-11
PSR-1685 Replace deprecated session-data endpoint

cyeniceri:


app/connectors/cache/SessionDataCacheConnector.scala
```diff
@@ -32,18 +32,17 @@ import javax.inject.Inject
...
     http
-      .get(url"${url(cacheId)}")
+      .get(url"$url")
       .execute[Option[SessionData]]
       .tapError(t => Future.successful(logger.warn(s"Failed to fetch $cacheId with message ${t.getMessage}")))
 
   override def remove(cacheId: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Unit] =
     http
-      .delete(url"${url(cacheId)}")
+      .delete(url"$url")
```
cyeniceri, line 45: @kberec @westwater @ccatosdevelopment
This remove method not used except spec file. Should we delete or use it when we log out ?

kberec: I don't think we can used it, as the session is shared with the other PODS applications. It would be better to remove it IMO.

app/connectors/cache/SessionDataCacheConnector.scala
```diff
@@ -32,18 +32,17 @@ import javax.inject.Inject
 class SessionDataCacheConnectorImpl @Inject()(config: FrontendAppConfig, http: HttpClientV2)
     extends SessionDataCacheConnector {
 
-  private def url(cacheId: String): String =
-    s"${config.pensionsAdministrator}/pension-administrator/journey-cache/session-data/$cacheId"
+  private def url: String = s"${config.pensionsAdministrator}/pension-administrator/journey-cache/session-data-self"
 
   override def fetch(cacheId: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Option[SessionData]] =
     http
```
kberec, line 38: Do you need to pass in some header? `loggedInAsPsa` maybe?

cyeniceri: Nope. I looked Pension Administrator, now they obtain the cacheId(externalId) in a same way as we used to (extract it from auth). Therefore, we don't need to send them anything.

app/connectors/cache/SessionDataCacheConnector.scala
cyeniceri: Let me delete this part of code

kberec (approved):

PR #1012, opened 2024-12-12
PSR-1679 | [Accessibility] Issue: Date fields validation

Neozxc:
**Issue:**
- [Date fields validation](https://github.com/hmrc/accessibility-audits-external/issues/8765)

**Dev Tasks:**
- [x] Only show the priority error

### DIFF
--- a/app/viewmodels/govuk/ErrorSummaryFluency.scala
+++ b/app/viewmodels/govuk/ErrorSummaryFluency.scala
@@ -41,7 +41,9 @@ trait ErrorSummaryFluency {
       errorLinkOverrides: Map[String, String]
     )(implicit messages: Messages): ErrorSummary = {
 
-      val allErrors = errors.distinctBy(_.message).map { error =>
+      val priorityError = errors.headOption.toList
+
+      val allErrors = priorityError.map { error =>
         ErrorLink(
           href = Some(s"#${errorLinkOverrides.getOrElse(error.key, error.key)}"),
           content = Text(messages(error.message, error.args.map {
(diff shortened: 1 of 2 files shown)


app/viewmodels/govuk/ErrorSummaryFluency.scala
```diff
@@ -41,7 +41,9 @@ trait ErrorSummaryFluency {
       errorLinkOverrides: Map[String, String]
     )(implicit messages: Messages): ErrorSummary = {
 
-      val allErrors = errors.distinctBy(_.message).map { error =>
+      val priorityError = errors.headOption.toList
```
kberec, line 44: This is common code. Are you sure this change doesn't break any functionality were we suppose to see multiple errors (e.g. there are multiple fields on the form).

app/viewmodels/govuk/ErrorSummaryFluency.scala
```diff
@@ -41,7 +41,9 @@ trait ErrorSummaryFluency {
...
+
```
kberec, line 45: Are you sure that it is always the first error that should be the priority one?

app/viewmodels/govuk/ErrorSummaryFluency.scala
kberec: Yes, but:

- Is the first one in `error` always the highest priority?
- In my understanding if there are multiple fields on a form and e.g. all have errors all should show one error. I.e. if there are 3 fields, the maximum number of errors shown on the error summary should be 3 (one for each field). I was just highlighting that this could potentially break that. I haven't checked.

cyeniceri:
This PR closed. Instead of this PR, we worked on https://github.com/hmrc/pension-scheme-return-frontend/pull/1016 and merged it.

PR #1019, opened 2024-12-16
PSR-1630 | Loans check and update page

ccatosdevelopment:


app/controllers/nonsipp/loansmadeoroutstanding/AmountOfTheLoanController.scala
```diff
@@ -108,7 +108,15 @@ class AmountOfTheLoanController @Inject()(
                 updatedAnswers <- Future
                   .fromTry(request.userAnswers.transformAndSet(AmountOfTheLoanPage(srn, index), value))
                 _ <- saveService.save(updatedAnswers)
-              } yield Redirect(navigator.nextPage(AmountOfTheLoanPage(srn, index), mode, updatedAnswers))
+              } yield
+                if (isPrePopulation && request.userAnswers.get(AreRepaymentsInstalmentsPage(srn, index)).isDefined) {
```
kberec, line 112: Potentially we could check here if the next cleared out page is still not defined instead of checking if the next retained pages is defined.
But as design is happy with this current approach after the demo today, we can leave it like this too.

kberec (approved):

app/controllers/nonsipp/loansmadeoroutstanding/AmountOfTheLoanController.scala
ccatosdevelopment: Thanks for the idea, that's a great shout! If we add that to the existing logic, then this should remove the `CheckMode` navigation edge case altogether. I think it would have to be added to the existing logic, rather than replacing it, otherwise newly added Loans would skip the `AreRepaymentsInstalmentsPage`. Since I'm working on the next Loans ticket at the moment, which is branched off of this branch, I'll make this addition as part of that ticket.

kberec: Yes please, but only if it is not too much of a complication.

PR #1022, opened 2024-12-17
Total term of loan field width #8777

Arei1997:


app/views/MultipleQuestionView.scala.html
```diff
@@ -93,7 +93,7 @@ <h1 class="govuk-heading-l govuk-!-margin-bottom-3">@renderMessage(viewModel.hea
         .withId(s"value.${index + 1}")
         .asNumeric()
         .withHint(field.hint.map(h => Hint(content = HtmlContent(renderMessage(h)))))
-        .withWidth(width.fold[InputWidth](InputWidth.Fixed10)(identity))
+        .withWidth(width.getOrElse(InputWidth.Fixed4))
```
kberec, line 96: Just to double check, would this default `Fixed4` work for all these where the value is not explicitly passed in:
image

Arei1997: I have checked, It works for the numeric only fields so other fields are not affected

kberec (approved):

PR #1026, opened 2024-12-19
PSR-1678 [Accessibility] Issue: Incorrect error summary link destinat…

cyeniceri:
…ion #8763 (P1)

### DIFF
--- a/app/views/ListRadiosView.scala.html
+++ b/app/views/ListRadiosView.scala.html
@@ -31,7 +31,7 @@
     @formHelper(action = viewmodel.onSubmit, Symbol("autocomplete") -> "off") {
 
         @if(form.hasErrors) {
-          @govukErrorSummary(ErrorSummaryViewModel(form))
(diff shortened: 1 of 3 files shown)


app/views/ListRadiosView.scala.html
```diff
@@ -31,7 +31,7 @@
     @formHelper(action = viewmodel.onSubmit, Symbol("autocomplete") -> "off") {
 
         @if(form.hasErrors) {
-          @govukErrorSummary(ErrorSummaryViewModel(form))
+          @govukErrorSummary(ErrorSummaryViewModel(form, errorLinkOverrides = Map("value" -> viewmodel.page.rows.head.index.toString)))
```
kberec, line 34: Can rows ever be empty?

cyeniceri: I don't think so, we should open this page only when there is data

cyeniceri: In case of empty list, it will show the whole component as it was.

Arei1997 (approved):

PR #1036, opened 2025-01-02
PSR-1688 autocomplete=off removal

Arei1997:


app/views/YesNoPageView.scala.html
```diff
@@ -30,7 +30,7 @@
 
 @layout(pageTitle = title(form, viewModel.title.toMessage), refresh = viewModel.refresh) {
 
-  @formHelper(action = viewModel.onSubmit, Symbol("autocomplete") -> "off") {
+  @formHelper(action = viewModel.onSubmit) {
```
kberec, line 33: @Arei1997 what about other occurrences of `autocomplete` `off`? Do we need to change everywhere else too?
As the ticket says "Journey/Step: Global"

kberec (approved):

PR #1041, opened 2025-01-06
Invalid HTML - duplicate attribute class

Arei1997:


app/views/TaskListView.scala.html
```diff
@@ -24,7 +24,7 @@
 @(viewmodel: PageViewModel[TaskListViewModel], schemeName: String)(implicit request: Request[_], messages: Messages)
 
 @item(content: Html) = {
-    <li class="govuk-task-list__item govuk-task-list__item--with-link">
```
kberec, line 27: I'm a bit confused with this change, as the ticket doesn't mention anything about a removal of `govuk-task-list__item--with-link` attribute.
It says: Remove class="govuk-task-list__link" from all task list link instances (keep class="govuk-link govuk-task-list__link"
But I cannot see that class duplicated in the application currently.
Is it possible that this was changed under a different ticket since the a11y testing?

Arei1997: I think that may be the case, I have been also looking. You are correct the removal should not have been govuk-task-list__item--with-link.

However when inspect elementing , "class="govuk-task-list__link"  has been removed and "class="govuk-link govuk-task-list__link" remains which seems to be in line with what they asked for 


This is what the href looks like now:

<a href="/pension-scheme-return/S2400000001/check-your-answers-basic-details" class="govuk-link govuk-task-list__link" aria-describedby="task-list-item-0-0">Change basic details</a>

Is this still incorrect do you think?

Vamshi-Muniganti: verified on main branch on tasklist normal and view pages and other pages, couldnt find any duplicate class mentioned this fix is really not needed, so closing this PR.

PR #1042, opened 2025-01-06
PSR-1670 | Add month names

Neozxc:
**Issue:**
- [Month field doesn’t accept month name](https://github.com/hmrc/accessibility-audits-external/issues/8753)

**Dev Tasks:**
- [x] Make sure month field accepts month names eg: "jan" "January"

### DIFF
--- a/test/forms/DateMappingsSpec.scala
+++ b/test/forms/DateMappingsSpec.scala
@@ -435,4 +435,25 @@ class DateMappingsSpec
       filledForm("value.year").value.value mustEqual date.getYear.toString
     }
   }
+
+  "must bind valid month data with short and full month names" in {
+
+    val validMonths = Table(
+      ("day", "month", "year", "expectedMonthValue"),
+      ("01", "Jan", "2020", 1),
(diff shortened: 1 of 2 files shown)


test/forms/DateMappingsSpec.scala
```diff
@@ -435,4 +435,25 @@ class DateMappingsSpec
       filledForm("value.year").value.value mustEqual date.getYear.toString
     }
   }
+
+  "must bind valid month data with short and full month names" in {
+
+    val validMonths = Table(
```
kberec, line 441: For completeness I would add a scenario where some invalid month textual values fail to bind. E.g.: `Febru`, `Anytext`.
But only do that if you think too that it would be useful.

kberec (dismissed):

test/forms/DateMappingsSpec.scala
Neozxc: This has been addressed

kberec (approved):

PR #1072, opened 2025-02-03
[PSR-1395] Add Check tag and verb

MadStu:


conf/messages.en
```diff
@@ -1125,97 +1125,119 @@ nonsipp.tasklist.history = View all versions of this return
 nonsipp.tasklist.description = You have completed {0} of {1} sections.
 nonsipp.tasklist.schemedetails.title = Scheme details
 nonsipp.tasklist.schemedetails.change.details.title = Change basic details
+nonsipp.tasklist.schemedetails.check.details.title = Check basic details
```
ccatosdevelopment, line 1128: Since only Member Details, Loans, Shares, Land or Property, Bonds, and Other Assets are pre-populated, they're the only sections that will require a `Check xyz` message in this file, so the rest can be safely removed.

app/utils/nonsipp/TaskListUtils.scala
```diff
@@ -530,8 +531,12 @@ object TaskListUtils {
           userAnswers.get(OtherAssetsCompleted.all(srn)).filter(_.nonEmpty),
           userAnswers.get(DidSchemeHoldAnySharesPage(srn))
         ) match {
-          case (Some(_), Some(_)) => List(quotedSharesItem, otherAssetsItem, otherAssetsDisposalItem)
-          case (Some(_), _) => List(otherAssetsItem, otherAssetsDisposalItem)
+          case (Some(_), Some(_)) =>
+            List(quotedSharesItem, otherAssetsItem) ++ (if
+              (quotedSharesStatusAndLink._1 != Check || otherAssetsStatus != Check) {
+              List(otherAssetsDisposalItem)} else { Nil })
```
ccatosdevelopment, line 537: I think this can be simplified, as Quoted Shares aren't pre-populated, so they can't have a status of `Check`, meaning we don't need a condition for that here. The rest of your logic still applies though, so I think the snippet below would work:

```
case (Some(_), Some(_)) =>
  List(quotedSharesItem, otherAssetsItem) ++
    (if (otherAssetsStatus != Check) List(otherAssetsDisposalItem) else Nil)
```

ccatosdevelopment (approved):

ccatosdevelopment (approved):

PR #1075, opened 2025-02-05
PSR-1634 Bonds pre pop check page

westwater:


app/controllers/nonsipp/bonds/UnregulatedOrConnectedBondsHeldCYAController.scala
```diff
@@ -120,18 +123,21 @@ class UnregulatedOrConnectedBondsHeldCYAController @Inject()(
...
-          fallbackCall =
-            controllers.nonsipp.bonds.routes.UnregulatedOrConnectedBondsHeldCYAController.onPageLoad(srn, index, mode)
-        )
-        .map {
-          case None => Redirect(controllers.routes.JourneyRecoveryController.onPageLoad())
-          case Some(_) =>
-            Redirect(navigator.nextPage(UnregulatedOrConnectedBondsHeldCYAPage(srn), NormalMode, request.userAnswers))
-        }
+      for {
+        updatedAnswers <- request.userAnswers.set(DidSchemeHoldAnySharesPage(srn), true).mapK
```
ccatosdevelopment, line 127: I think a lil copy & paste error slipped in here!

westwater: we need to update the initial page here so when we navigate to the list page next, we don't bounce back into the journey - I'm not 100% sure why it doesn't work when I do this in the bonds preprocessor but this is present in all CYA pages

ccatosdevelopment: You're absolutely right, but you've used the first question page from Shares, rather than Bonds, that's all!

westwater: 🤦 good spot!!! thanks for that

ccatosdevelopment (approved):

app/controllers/nonsipp/bonds/UnregulatedOrConnectedBondsHeldCYAController.scala
ccatosdevelopment: No worries, it happens to the best of us! 😄

PR #1080, opened 2025-02-07
PSR-1786: update TaskListController and LoansListController behaviour…

jamalosman:
… to support pre-pop flag in loanTransactions.

### DIFF
--- a/app/pages/nonsipp/loansmadeoroutstanding/LoanPrePopulated.scala
+++ b/app/pages/nonsipp/loansmadeoroutstanding/LoanPrePopulated.scala
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2025 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package pages.nonsipp.loansmadeoroutstanding
+
+import utils.RefinedUtils.RefinedIntOps
+import pages.QuestionPage
+import config.RefinedTypes.Max5000
+import models.SchemeId.Srn
+import play.api.libs.json.JsPath
+
+case class LoanPrePopulated(srn: Srn, index: Max5000) extends QuestionPage[Boolean] {
+  override def path: JsPath = Paths.loans \ toString \ index.arrayIndex.toString
+  override def toString: String = "loanPrePopulated"
+}
--- a/app/prepop/LoansPrePopulationProcessor.scala
+++ b/app/prepop/LoansPrePopulationProcessor.scala
@@ -17,18 +17,24 @@
 package prepop
 
 import pages.nonsipp.loansmadeoroutstanding.Paths.loans
-import models.UserAnswers.SensitiveJsObject
+import config.RefinedTypes.Max5000
 import models.SchemeId.Srn
-import play.api.libs.json._
 import models.UserAnswers
+import pages.nonsipp.common.LoanIdentityTypePages
 import pages.nonsipp.loansmadeoroutstanding._
+import utils.ListUtils.ListOps
+import models.UserAnswers.SensitiveJsObject
+import play.api.Logger
+import play.api.libs.json._
 
-import scala.util.{Success, Try}
+import scala.util.Try
 
 import javax.inject.{Inject, Singleton}
 
 @Singleton
 class LoansPrePopulationProcessor @Inject()() {
+
+  private val logger = Logger(getClass)
 
   def clean(baseUA: UserAnswers, currentUA: UserAnswers)(srn: Srn): Try[UserAnswers] = {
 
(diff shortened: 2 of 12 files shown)


app/pages/nonsipp/loansmadeoroutstanding/LoanPrePopulated.scala
```diff
@@ -0,0 +1,35 @@
...
+  override def path: JsPath = Paths.loans \ toString \ index.arrayIndex.toString
+  override def toString: String = "loanPrePopulated"
+}
+
+object LoanPrePopulated {
+  def all(srn: Srn): IndexedQuestionPage[Boolean] = new IndexedQuestionPage[Boolean] {
+    override def path: JsPath = Paths.loans \ toString
+    override def toString: String = "loanPrePopulated"
+  }
+}
```
ccatosdevelopment, line 35: I noticed this doesn't have any usages - is there a specific purpose that we're planning on using this for? If not, I'd suggest removing it.

app/transformations/LoansTransformer.scala
```diff
@@ -510,6 +511,16 @@ class LoansTransformer @Inject() extends Transformer {
...
       )
 
+      loanPrePopulated = indexes
+        .filter(
+          index => {
+            loanTransactions(index.value - 1).prePopulated.nonEmpty
+          }
+        )
+        .map(
+          index => LoanPrePopulated(srn, index) -> loanTransactions(index.value - 1).prePopulated.getOrElse(false)
```
ccatosdevelopment, line 521: Another small suggestion - since we're filtering `prePopulated` fields that are `nonEmpty` in line 517, I think it would make sense to substitute the `.getOrElse(false)` with a `.get`, which would make this more consistent with the rest of this file's implementation.

app/utils/nonsipp/check/LoansCheckStatusUtils.scala
```diff
@@ -71,35 +71,12 @@ object LoansCheckStatusUtils {
     userAnswers: UserAnswers,
```
ccatosdevelopment, line 71: Since the implementation of this method has been changed, the documentation isn't entirely accurate anymore, so I'd suggest updating the second sentence to something like: "A Loans record only needs to be checked if its LoanPrePopulated field is false."

app/utils/nonsipp/check/LoansCheckStatusUtils.scala
```diff
@@ -71,35 +71,12 @@ object LoansCheckStatusUtils {
...
+      case Some(checked) => !checked
+      case None => false // non-pre-pop
     }
 
-    anyPrePopClearedAnswersMissing && allOtherAnswersPresent
-  }
-
   /**
    * This method determines whether or not all answers are present for a given IdentityType.
    * @param userAnswers the answers provided by the user
```
ccatosdevelopment, line 82: Since this method is no longer used, my suggestion would be to delete it.

test/utils/nonsipp/check/LoansCheckStatusUtilsSpec.scala
```diff
@@ -34,8 +34,11 @@ class LoansCheckStatusUtilsSpec extends ControllerBaseSpec with Matchers with Op
   private val conditionalYesSecurity: ConditionalYes[Security] = ConditionalYesNo.yes(security)
   private val conditionalYesArrears: ConditionalYes[Money] = ConditionalYesNo.yes(money)
 
-  private val schemeHadLoansTrue = defaultUserAnswers.unsafeSet(LoansMadeOrOutstandingPage(srn), true)
-  private val schemeHadLoansFalse = defaultUserAnswers.unsafeSet(LoansMadeOrOutstandingPage(srn), false)
+  private val schemeHadLoansTrue = defaultUserAnswers
```
ccatosdevelopment, line 37: Because the implementation of the Check method has been drastically simplified, most of the tests and test data in this file can also be simplified, or deleted entirely - in the interest of time, I think I'll draft up a new implementation for this Spec, and we can go through it together, then implement it here, if that's alright with you.

jamalosman: sure

ccatosdevelopment (approved):

app/utils/nonsipp/check/LoansCheckStatusUtils.scala
```diff
@@ -71,84 +69,9 @@ object LoansCheckStatusUtils {
     userAnswers: UserAnswers,
     srn: Srn,
     recordIndex: Max5000
-  ): Boolean = {
```
kberec, line 74: Do we want to keep the existing logic as a fallback?
I'm thinking about the scenario that this change is rolled out to QA environment BEFORE ETMP changes are rolled out. I.e. prePopulation flag introduced and values we are passing in retained). If we remove this logic, there is a chance something will be broken in this transition period.

jamalosman: I see... would we be able to detect if the changes are in ETMP? or would we use a feature flag maybe?

kberec: We would need to remove the fallback logic later on. On or after 11th March -  this is when we expect these  ETMP changes to be rolled out.

ccatosdevelopment (changes requested):
Requesting changes due to additional AC being identified.

ccatosdevelopment (approved):

app/prepop/LoansPrePopulationProcessor.scala
```diff
@@ -48,7 +51,24 @@ class LoansPrePopulationProcessor @Inject()() {
 
     cleanUpOptionalFields(transformedResult, indexesToDelete) match {
       case JsSuccess(value, _) =>
-        Success(currentUA.copy(data = SensitiveJsObject(value.deepMerge(currentUA.data.decryptedValue))))
+        val uaWithLoansData = currentUA.copy(data = SensitiveJsObject(value.deepMerge(currentUA.data.decryptedValue)))
+
+// TODO: once ETMP changes are in QA then we need to un comment this code
```
kberec, line 56: What happens if we un-comment it now?

jamalosman: We will send the pre-pop flag to ETMP

kberec (dismissed):

PR #1082, opened 2025-02-10
[PSR-1692] Bug Fix - Disallow new lines in FE Validation

MadStu:


app/models/Security.scala
```diff
@@ -34,7 +34,7 @@ object Security extends (String => Security) {
   implicit val securityWrite: Writes[Security] = new SimpleObjectWrites[Security](_.value)
   implicit val securityRead: Reads[Security] = new SimpleObjectReads[Security]("security", Security.apply)
 
-  private val validSecurityFormat = """^[a-zA-Z0-9\-'" \t\r\n,.@#/]+$"""
```
kberec, line 37: On part of the regex is changing from `#/` to `/&()`. 
Do you know why was it different than the regex in TextFormProvider?
If no, should we keep `#/`?

test/services/MemberDetailsUploadValidatorSpec.scala
```diff
@@ -253,7 +253,7 @@ class MemberDetailsUploadValidatorSpec extends BaseSpec with TestValues {
             ValidationError("F3", DateOfBirth, "memberDetails.dateOfBirth.upload.error.future"),
             ValidationError("C4", LastName, "memberDetails.lastName.upload.error.invalid"),
             ValidationError("F5", DateOfBirth, "memberDetails.dateOfBirth.upload.error.future"),
-            ValidationError("E5", NoNinoReason, "noNINO.upload.upload.error.invalid"),
+            ValidationError("E5", NoNinoReason, "error.textarea.invalid"),
```
kberec, line 256: In this case, the validation is agains a value in the CSV file and not against a value of the text field on the UI. It would be worthwhile to ask design / content design if they are still happy with the error message "The submitted text .... " here.

app/models/Security.scala
MadStu: I don't know the reason why it uses it's own regex validation but the only place it's used is for the description of the loan security item, so it's not like a password kind of security, just a description of whatever thing the loan is secured against.
Previously the invalid character error message for this was.:

```securityGivenForLoan.securityGiven.yes.conditional.error.invalid = Details of the security must only include letters a to z, numbers 0 to 9, apostrophes, ‘atʼ signs (@), commas, forward slashes, full stops, hashes (#), hyphens, speech marks, ampersands, brackets and spaces```

I updated the regex so they're all the same for consistency but happy to change back. If ETMP allows hashes then maybe that should also be updated for the other regex validations?

kberec (approved):

PR #1087, opened 2025-02-17
PSR-1842 Added numeric sorting to address lookup

kberec:


app/services/AddressService.scala
```diff
@@ -30,9 +31,71 @@ class AddressService @Inject()(connector: AddressLookupConnector)(
...
-  private def addressFromALFAddress(lookupResponse: ALFAddressResponse): Address =
+  private def sortAddresses(unsorted: List[Address]): List[Address] =
+    unsorted.sortBy(
+      s =>
+        (
+          s.street.getOrElse(""),
+          s.houseNumber.getOrElse(Int.MaxValue),
+          s.flatNumber.getOrElse(Int.MaxValue),
+          s.flat.getOrElse("")
+        )
```
westwater, line 44: How does tuple sorting work here? does it sort downward so if 2 people are on the same street (as it's the first in the tuple) it will then sort based on the house number then flat number etc? Not sure about the Int max values though - might be worth a comment

kberec: Yes, it will take street as the first criteria of the sorting, then, houseNumber if present, then flatNumber if present or the remaining flat description.
If any of these not present, default value (e.g. Int.MaxValue) will push the item to the last position in that criteria-group.

westwater (approved):

PR #1101, opened 2025-02-21
[PSR-1856] Bug Fix Counting spaces in textarea character limits

MadStu:


app/forms/mappings/Mappings.scala
```diff
@@ -219,7 +219,7 @@ trait Mappings extends Formatters with Constraints {
         case (mapping, (regex, key)) =>
           mapping.verifying(verify[String](key, _.filterNot(_.isWhitespace).matches(regex), args: _*))
       }
-      .verifying(verify[String](maxLengthErrorKey, _.filterNot(_.isWhitespace).length <= maxLength, args: _*))
+      .verifying(verify[String](maxLengthErrorKey, _.length <= maxLength, args: _*))
```
kberec, line 222: The same logic is present below in `nonRequiredPostCode` method (used in land or property uk manual entry address page). It might not be a problem like this due to the additional transformation happening on the postcoode, but it might be worthwhile to double check.

app/forms/mappings/Mappings.scala
```diff
@@ -219,7 +219,7 @@ trait Mappings extends Formatters with Constraints {
...
```
kberec, line 223: A unit test could be useful to prove whitespace characters count towards the maximum limit.

app/forms/mappings/Mappings.scala
MadStu: I've just double checked `nonRequiredPostCode` and it's working as it should

PR #1102, opened 2025-02-24
PSR-1819: update members list page for prepop check scenario

jamalosman:


app/controllers/nonsipp/memberdetails/SchemeMembersListController.scala
```diff
@@ -138,7 +141,8 @@ class SchemeMembersListController @Inject()(
                 optCurrentVersion = request.currentVersion,
                 optPreviousVersion = request.previousVersion,
                 compilationOrSubmissionDate = request.userAnswers.get(CompilationOrSubmissionDatePage(srn)),
-                showBackLink = showBackLink
+                showBackLink = showBackLink,
+                prePopNotChecked = request.userAnswers.get(MembersDetailsChecked(srn)).contains(false)
```
kberec, line 145: It might be an overkill, but potentially you can utilise the existing `isPrePopulation` method as an additional guard condition here, e.g. something like:
`` prePopNotChecked = isPrePopulation && request.userAnswers.get(MembersDetailsChecked(srn)).contains(false) ``
to make sure this applies for pre-population only. Even if in theory for not pre-populated records this should never be false, it will add some additional certainty IMO.

app/controllers/nonsipp/memberdetails/SchemeMembersListController.scala
```diff
@@ -180,12 +184,33 @@ class SchemeMembersListController @Inject()(
               }
           }.merge.pure[Future],
           value => {
-            if (lengthOfMembersDetails == maxSchemeMembers && value) {
+            val notChecked = request.userAnswers.get(MembersDetailsChecked(srn)).contains(false)
+
+            if (notChecked) {
```
kberec, line 189: Same to consider here as in above comment. E.g. 
`` if (isPrePopulation && notChecked) ``

app/controllers/nonsipp/memberdetails/SchemeMembersListController.scala
```diff
@@ -180,12 +184,33 @@ class SchemeMembersListController @Inject()(
...
+                updatedAnswers <- Future.fromTry(
+                  request.userAnswers
+                    .set(MembersDetailsChecked(srn), true)
+                )
+                _ <- saveService.save(updatedAnswers)
+                _ <- psrSubmissionService
+                  .submitPsrDetailsWithUA(
+                    srn,
+                    updatedAnswers,
+                    fallbackCall = routes.SchemeMembersListController.onPageLoad(srn, page, manualOrUpload)
```
kberec, line 200: This fallback call is where the error page should navigate when the submission (POST) fails.
Would it be better to navigate to the task list?
`fallbackCall = controllers.nonsipp.routes.TaskListController.onPageLoad(srn)`

kberec (dismissed):

kberec (approved):

PR #1104, opened 2025-02-25
PSR-1420 if members are more than 99 nav to BasicCYA page

narendrapaduchuri:


app/controllers/OverviewController.scala
```diff
@@ -325,12 +325,16 @@ class OverviewController @Inject()(
           }
 
         case _ =>
-          Future.successful(
-            Redirect(controllers.nonsipp.routes.TaskListController.onPageLoad(srn))
-              .addingToSession(
-                Constants.PREPOPULATION_FLAG -> String.valueOf(lastSubmittedPsrFbInPreviousYears.isDefined)
-              )
-          )
+          val byPassedJourney =
```
kberec, line 328: Do we need some tests for this?

PR #1105, opened 2025-03-04
[PSR-1826] Accessible name implementation on table wrapper

MadStu:
[Accessibility] Issue: Accessible name implementation on table wrapper (P1) #9034


rameshmandanapu1:
image

PR #1119, opened 2025-03-20
Pre-population transform to ETMP with top level fields

kberec:


kberec:
Replaced by https://github.com/hmrc/pension-scheme-return-frontend/pull/1120

PR #1126, opened 2025-03-26
[PSR-1621] Increase IF Timeout

MadStu:


app/connectors/MinimalDetailsConnector.scala
```diff
@@ -42,6 +42,7 @@ class MinimalDetailsConnector @Inject()(appConfig: FrontendAppConfig, http: Http
     http
       .get(url"$url")
       .setHeader("loggedInAsPsa" -> String.valueOf(loggedInAsPsa))
+      .transform(_.withRequestTimeout(appConfig.ifsTimeout))
```
kberec, line 45: Unless there is a specific reason, I don't think it is needed here, as this is a different service (not IF).

kberec (approved):

PR #1138, opened 2025-05-01
PSR-1927 What you will need pages content updated

kberec:


kberec (commented):
Changes look good, but I cannot approve as co-authored.

jamalosman (approved):

PR #1156, opened 2025-05-21
[PSR-1751] Shares Disposal Smart Navigation

MadStu:


jamalosman:
I think you need one more change in `ReportedSharesDisposalListController.getCompletedDisposals` to filter out non-completed disposals

image

Also not sure what's up with this 'temporary E2E workaround'. Looks strange

image

Other than that, it looks good to me.

kberec (approved):

PR #1184, opened 2025-06-27
[PSR-1461] Increase Unit Test Coverage

MadStu:


app/navigation/nonsipp/LandOrPropertyNavigator.scala
```diff
@@ -276,10 +274,10 @@ object LandOrPropertyNavigator extends JourneyNavigator {
         case page @ IsLandPropertyLeasedPage(srn, index) =>
           if (userAnswers.get(page).contains(true)) {
             controllers.nonsipp.landorproperty.routes.LandOrPropertyLeaseDetailsController
-              .onPageLoad(srn, index, NormalMode)
+              .onPageLoad(srn, index, CheckMode)
```
kberec, line 277: Are you sure you need to amend the logic here?
In some cases there was an intentional mode change in navigation, i.e. navigating from a page in change mode to a page in normal mode and vice versa.
I'm not sure if this here was intentional or not.

kberec (approved):

PR #1199, opened 2025-07-08
[PSR-2058] Fix Loan not getting listed when zero value bug

MadStu:


app/controllers/nonsipp/loansmadeoroutstanding/LoansListController.scala
```diff
@@ -338,7 +338,7 @@ object LoansListController {
         List()
       case (list, _) =>
         list.flatMap { case LoansData(index, loanAmount, recipientName, loansProgress, canRemove) =>
-          val validLoan = loanAmount.value > 0 && recipientName.nonEmpty
+          val validLoan = recipientName.nonEmpty
```
kberec, line 341: Because of the `loanAmount.value > 0` removal and `validLoan`'s usage further down the line, e.g. in 
  `case _ if loansProgress.inProgress && !isPrePopulation => None`
  I suggest to do some extra testing of the rendering of this list page. E.g. in pre-population mode or anywhere else the display logic can be affected.

kberec (approved):

PR #1238, opened 2025-08-07
Psr 2006 2

jamalosman:


app/controllers/nonsipp/declaration/SummaryController.scala
```diff
@@ -128,39 +134,40 @@ class SummaryController @Inject() (
...
           psrVersionsService
             .getVersions(request.schemeDetails.pstr, startDate, srn)
         )
-        submittedVersions = versions
-          .filter(_.reportStatus == SubmittedAndSuccessfullyProcessed)
-        lastSubmitted = submittedVersions
-          .maxByOption(_.compilationOrSubmissionDate)
-          .getOrElse(versions.maxBy(_.compilationOrSubmissionDate))
-        lastSubmittedDateTime = lastSubmitted.compilationOrSubmissionDate
+        lastVersion = versions
```
kberec, line 152: I think we should go with the highest version number here and not the date of submission. In theory it should be the same, but some other logic in the service also sorts / searches by the highest version.

test-utils/generators/ModelGenerators.scala
```diff
@@ -90,7 +90,7 @@ trait ModelGenerators extends BasicGenerators {
       status <- schemeStatusGen
       schemeType <- nonEmptyString
       authorisingPsa <- Gen.option(nonEmptyString)
-      establishers <- Gen.listOf(establisherGen)
+      establishers <- Gen.listOfN(5, establisherGen)
```
kberec, line 93: 👍

kberec (approved):

PR #1258, opened 2025-09-03
PSR-2125: add new flag to submitted audit event

jamalosman:


app/models/audit/PSRSubmissionAuditEvent.scala
```diff
@@ -45,7 +45,8 @@ case class PSRSubmissionAuditEvent(
       "credentialRolePsaPsp" -> credentialRole,
       "taxYear" -> taxYear.toYearFormat,
       "date" -> LocalDate.now(ZoneId.of("Europe/London")).format(DateTimeFormatter.ofPattern("dd MMMM yyyy")),
-      "payload" -> Json.toJson(psrSubmission)
+      "payload" -> Json.toJson(psrSubmission),
+      "submissionAmendment" -> psrSubmission.minimalRequiredSubmission.reportDetails.fbVersion.exists(_ > "001")
```
kberec, line 49: Does this need to compare with int, i.e. `.toInt() > 1`?

jamalosman: ordinal string comparison should work okay, but we can convert to int

MadStu (approved):

PR #1266, opened 2025-09-18
[PSR-2126] Change to only allow one audit start event per return

MadStu:
Add Tests

### DIFF
--- a/test/controllers/nonsipp/BasicDetailsCheckYourAnswersControllerSpec.scala
+++ b/test/controllers/nonsipp/BasicDetailsCheckYourAnswersControllerSpec.scala
@@ -23,6 +23,7 @@ import utils.nonsipp.summary.BasicDetailsCheckAnswersUtils
 import play.api.inject.bind
 import utils.IntUtils.given
 import cats.implicits.toShow
+import config.Constants.CIP_START_EVENT_FLAG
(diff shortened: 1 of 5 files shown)


test/controllers/nonsipp/BasicDetailsCheckYourAnswersControllerSpec.scala
```diff
@@ -350,6 +354,30 @@ class BasicDetailsCheckYourAnswersControllerSpec
...
+          mockTaxYear(currentReturnTaxYear)
+          MockPsrSubmissionService.submitPsrDetails()
+
+          val request = FakeRequest(GET, onSubmit.url).withSession(CIP_START_EVENT_FLAG -> "true")
+          val result = route(application, request).value
+          status(result) mustEqual SEE_OTHER
+          verify(mockAuditService, times(1)).sendEvent(any())(using any(), any())
+          reset(mockAuditService)
+
+          val requestWithoutFlag = FakeRequest(GET, onSubmit.url)
```
kberec, line 372: Very minor: there might be two tests here:
1. "should send audit event when CIP_START_EVENT_FLAG is true"
2. "should not send audit event when CIP_START_EVENT_FLAG is not present/defined"

